PR #930, opened 2013-11-03
Enable shumway in private browsing window.

saneyuki:
At the current implementation, Shumway doesn't run in private browsing window on Firefox even if we set `shumway.disabled` as `false`.  I think that this is not good for testing for feedbacks, and this behavior confuse us for  coherency of the configuration.


yurydelendik:
There is the shumway.enableForPrivate about:config setting to enable that for private mode. See https://github.com/mozilla/shumway/wiki/Debugging-and-Configuring-Shumway . Thanks

PR #1320, opened 2014-05-26
Add disableAvm flag, which turns off both AVM1 and AVM2.

laszlopandy:
This is extremely useful for anyone who wants to host and display a SWF on their site, but does not know or does not trust the author of the SWF.

Adding `disableAvm: true` to the `SWF.embed()` options allows SWFs to render and play, but without executing any of the code inside (which could potentially be malicious).


mbebenita:
This PR is stale, a lot has changed. If we still need this feature, we should submit a new PR.

PR #1321, opened 2014-05-26
Fontfixes

rszaloki:
Improves font and label rendering.

If a swf generated by a converter (eg: pdf to swf), then sometimes the converter cannot create a valid font information. Also the "text" label shouldn't flip vertically, but the text2 should.


mbebenita:
@tobytailor are there fixes here we still need to cherrypick?

tobytailor:
Nope, everything should already be fixed in master.

PR #1968, opened 2014-12-17
Fixes extension/iframe font loading regression.

yurydelendik:


tschneidereit:
I already had a WIP patch that did more than this one in flight, and have landed that in #1969.

PR #1969, opened 2014-12-17
Bug 1104270 - fix eager font and image parsing

tschneidereit:


src/gfx/renderables/renderables.ts
```diff
@@ -375,40 +376,41 @@ module Shumway.GFX {
...
-        image.onerror = function () {
-          unexpected("Image loading error: " + ImageType[type]);
-        };
+      if (type === ImageType.JPEG || type === ImageType.PNG || type === ImageType.GIF) {
+        release || Debug.assertUnreachable("Mustn't encounter un-decoded images here");
       } else {
+        var bounds = this._bounds;
+        var imageData = this._imageData;
+        if (!imageData || imageData.width !== bounds.w || imageData.height !== bounds.h) {
+          imageData = new ImageData(new Uint8ClampedArray(dataBuffer.length), bounds.w, bounds.h);
```
tschneidereit: Hmm, I don't get this error, neither on release nor in Canary. Perhaps it's behind some experimental features flag?

yurydelendik: how is that different from `imageData = this._context.createImageData(bounds.w, bounds.h);`

tschneidereit:
> I created an additional test db~/images/loading.swf that uses Flash's pre-loader template. Most of the time I never see the pre-loader, it just hangs until the images are displayed. When I do see the pre-loader, Images are not displayed.

Thanks, that and `images.swf` were incredibly helpful in finding various bugs. With the additional patches, both of the test SWFs work flawlessly in Nightly and Canary. Speed is quite good, too, with CPU usage roughly matching the Flash Player's.

src/gfx/renderables/renderables.ts
tschneidereit: It's not really different, just slightly less ugly. I'll change it to improve compatibility, though.

tschneidereit: Well, actually it is different in that fewer temporaries are needed. Still changing it.

tschneidereit: (And of course the previous comment is wrong and this is just a-ok.)

tschneidereit:
> Most of the time I never see the pre-loader, it just hangs until the images are displayed.
This is actually true in Flash, too. It's just an artifact of the loading speed being dominated by CPU, not network. If you were to load this SWF from the network, you'd see the preloader.

That being said, I think we should implement [bug 1112363](https://bugzilla.mozilla.org/show_bug.cgi?id=1112363), which would still cause us to show the preloader in this case, very likely.

src/gfx/remotingGfx.ts
```diff
@@ -182,23 +170,39 @@ module Shumway.Remoting.GFX {
...
       var image = new Image();
-      var self = this;
+      var asset = RenderableBitmap.FromImage(image, -1, -1);
       image.src = URL.createObjectURL(new Blob([data], {type: getMIMETypeForImageType(type)}));
       image.onload = function () {
-        self._canvas.width = image.width;
-        self._canvas.height = image.height;
-        self._context.drawImage(image, 0, 0);
-        oncomplete(self._context.getImageData(0, 0, image.width, image.height));
+        asset.setBounds(new Rectangle(0, 0, image.width, image.height));
```
tschneidereit: Ok, done.

yurydelendik:
Looks good.

PR #1973, opened 2014-12-26
Remove redundant acid tests and speed up some of the remaining ones

tschneidereit:


tschneidereit:
Surprisingly small win, but I'll take it.

PR #1977, opened 2014-12-30
Bug 1116467 - Beef up E4X compatibility

tschneidereit:


src/avm2/natives/xml.ts
```diff
@@ -71,28 +71,34 @@ module Shumway.AVM2.AS {
...
+        if (node.hasSimpleContent()) {
+          var s = '';
+          for (var i = 0; i < node._children.length; i++) {
+            var child = node._children[i];
+            if (child._kind === ASXMLKind.Comment ||
+                child._kind === ASXMLKind.ProcessingInstruction)
+            {
+              continue;
+            }
+            s += toString(child);
```
yurydelendik: probably not a big deal here, but shall we set a pattern to not use `+=` for strings in loops? in many cases it is further optimized to use array.push/.join('') later by somebody else.

src/avm2/natives/xml.ts
```diff
@@ -347,82 +353,83 @@ module Shumway.AVM2.AS {
...
+    if (value === undefined) {
+      Runtime.throwError('TypeError', Errors.ConvertUndefinedToObjectError);
+    }
+    if (value instanceof ASXML) {
+      targetList.append(value);
+      return;
+    }
+    // The E4X spec says we must throw a TypeError for non-Boolean, Number, or String objects.
+    // Flash thinks otherwise.
+    var parentString = '<parent xmlns="' + ASXML.defaultNamespace + '">' + value + '</parent>';
```
yurydelendik: it might was wrong before, but we shall escapeAttributeValue for `ASXML.defaultNamespace`

src/avm2/natives/xml.ts
```diff
@@ -1238,6 +1239,7 @@ module Shumway.AVM2.AS {
   export class ASXML extends ASNative implements XMLType {
     public static instanceConstructor: any = ASXML;
     static classInitializer: any = function() {
+      AS.ASXML = this;
```
yurydelendik: This one needs some comment -- this one confuses me :)

yurydelendik: Shall we have one in the ASXMLList?

yurydelendik:
Looks good if AVM2 tests pass.

/botio test

yurydelendik:
@mbebenita  do we test only the AVM2 interpreter in test_avm2_quick ? do you see anything that can break the compiler execution?

src/avm2/natives/xml.ts
tschneidereit: I don't think such a general rule makes much sense. There's a size/speed tradeoff to be made, plus it's not even entirely clear that the push/join pattern is better when Ion is active.

Also, this all is *so* not fast, so even if there's a speed difference to be had here, it just doesn't matter in the grand scheme of things.

src/avm2/natives/xml.ts
tschneidereit: Good call, fixed.

src/avm2/natives/xml.ts
tschneidereit: It really does need a comment - I don't understand it anymore myself ...

tschneidereit:
> @mbebenita do we test only the AVM2 interpreter in test_avm2_quick ?

We should go through and enable all the Tamarin tests that we currently pass. If we don't run them in `reftest`, the slowdown that'll introduce seems acceptable.

> do you see anything that can break the compiler execution?

I tested with both compiler and interpreter locally.

PR #1978, opened 2014-12-30
Bug 1116468 - Improve describeType compatibility

tschneidereit:


yurydelendik:
The main issue I see is creating XML as a string. In this case it may not harm anybody, but making e.g. `"<a b='" + n + "' />"` where `n = "'>test<c d='"`, might introduce the unexpected behavior and exceptions. We need to safe operation, similar that was used is the AS code. Also, AS.ASXML('<tag/>'); is costly, do we have an operation to create an element with specified name without parsing the string?

tschneidereit:
> The main issue I see is creating XML as a string. In this case it may not harm anybody, but making e.g. "<a b='" + n + "' />" where n = "'>test<c d='", might introduce the unexpected behavior and exceptions. We need to safe operation, similar that was used is the AS code.

At first I wanted to reply "this is safe because it operates on controlled input", but then I realized that a malicious SWF could well have class names that break this. To what end, I don't know.

> Also, AS.ASXML(''); is costly, do we have an operation to create an element with specified name without parsing the string?

Yes, we should have special-casing for at least empty strings in ASXML. But then again, the entire E4X stuff is ridiculously inefficient, so this probably matters frighteningly little.

PR #2006, opened 2015-01-14
Bug 1121538 - Add AVM2 execution mode that only compiles hot functions

tschneidereit:
This introduces ExecutionMode.COMPILE_HOT, which lets all functions run in the interpreter until they get hot enough. (Where "hot enough" is 500 times, for now, but we can do much better by measuring time spent in the function or how many opcodes we executed on average or something like that.)

### DIFF
--- a/src/flash/system/SecurityDomain.as
+++ b/src/flash/system/SecurityDomain.as
@@ -19,7 +19,6 @@ package flash.system {
 [native(cls='SecurityDomainClass')]
 public class SecurityDomain {
   public static native function get currentDomain():SecurityDomain;
-  public function SecurityDomain() {}
-  public native function get domainID():String;
+  public native function SecurityDomain();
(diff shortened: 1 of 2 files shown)


tschneidereit:
Ok, so I benchmarked this with the v8 stuff and sadly, it completely destroys performance. (As in, it's hundreds of times slower.)

I also have a prototype implementation of a `replaceFunctionScript` intrinsic, but it doesn't work with the JITs enabled. Might be that it can be made to work simply by invalidating the JITed code, but I don't yet know. I also don't know if that'd completely restore performance.

tschneidereit:
(Since this is disabled by default, I still think we should land it.)

tschneidereit:
This won't land in the current form anyway, and I accidentally pushed it from master. So, closing.

PR #2010, opened 2015-01-16
Implement support for CSS formated text fields

tobytailor:


src/TextContent.ts
```diff
@@ -133,7 +133,7 @@ module Shumway {
       this.coords = null;
     }
 
-    parseHtml(htmlText: string, multiline: boolean = false) {
+    parseHtml(htmlText: string, styleSheet: flash.text.StyleSheet = null, multiline: boolean = false) {
```
tschneidereit: No default values if not really necessary please.

src/avm2/native.ts
```diff
@@ -246,6 +246,29 @@ module Shumway.AVM2.AS {
       }
       return Shumway.StringUtilities.concat3("[object ", self.class.classInfo.instanceInfo.name.name, "]");
     }
+
+    /**
+     * Transforms an AS value into a JS value.
```
tschneidereit: Make this comment say what that actually means.

src/flash/text/StyleSheet.as
```diff
@@ -17,26 +17,15 @@
...
-  public function transform(formatObject: Object): TextFormat {
-    notImplemented("transform");
-    return null;
-  }
-  public function parseCSS(CSSText: String): void { notImplemented("parseCSS"); }
+  public native function StyleSheet();
+  public native function get styleNames(): Array;
+  public native function getStyle(styleName: String): Object;
+  public native function setStyle(styleName: String,
+                                  styleObject: Object): void;
```
tschneidereit: uber-nit: doesn't need to be wrapped.

src/flash/text/StyleSheet.ts
```diff
@@ -18,37 +18,152 @@ module Shumway.AVM2.AS.flash.text {
...
+      return names;
+    }
+
+    getStyle(styleName: string): Style {
+      styleName = asCoerceString(styleName);
+      var style = this._rules[styleName.toLowerCase()];
+      if (!style) {
+        return null;
+      }
+      return Shumway.ObjectUtilities.cloneObject(style);
```
tschneidereit: This needs to convert back to an `ASObject` instead of just cloning.

src/flash/text/StyleSheet.ts
```diff
@@ -18,37 +18,152 @@ module Shumway.AVM2.AS.flash.text {
...
+    }
+
+    setStyle(styleName: string, styleObject: Style) {
+      styleName = asCoerceString(styleName);
+      this._rules[styleName.toLowerCase()] = Shumway.AVM2.AS.ASObject.transformASValueToJS(styleObject);
```
tschneidereit: The function name transformASValueToJS sounds like it would deeply convert the value. What it really does (and what we actually want, here, I think) is a shallow clone with only public names, stripped of their qualifiers. Perhaps call it something like `shallowClone`, with `stripQualifiers` as a parameter? Or (because I admit we probably always want the qualifiers stripped) `transferPublicPropertiesToJSObject`, with the target object passed in?

(Also, no need to use such a long path to the function: you can access all names in namespaces above the current one without qualifying them. I.e., in this case `ASObject.transformASValueToJS` would suffice.)

src/flash/text/StyleSheet.ts
```diff
@@ -18,37 +18,152 @@ module Shumway.AVM2.AS.flash.text {
...
+      styleName = asCoerceString(styleName);
+      var style = this._rules[styleName.toLowerCase()];
+      if (!style) {
+        return null;
+      }
+      return Shumway.ObjectUtilities.cloneObject(style);
+    }
+
+    setStyle(styleName: string, styleObject: Style) {
+      styleName = asCoerceString(styleName);
```
tschneidereit: I just checked: only objects and arrays are stored, so we need to bail out early if `typeof styleObject !== 'object'`.

src/TextContent.ts
```diff
@@ -161,6 +161,16 @@ module Shumway {
           }
         },
         start: (tagName, attributes) => {
+          var style = null;
+          if (styleSheet) {
+            style = styleSheet.getStyle(tagName);
```
tschneidereit: `getStyle` is a hugely expensive operation, and we end up essentially reverting everything it does in `textFormat.merge` right below. How about we instead introduce an internal `StyleSheet#_getRawStyle` function that returns the object unmodified, and then have a `TextFormat#mergeStyleDefinition` method to use these? That'd just need iterate over the properties on the style definition and copy those over that are valid format properties.

src/flash/text/StyleSheet.ts
```diff
@@ -18,37 +18,152 @@ module Shumway.AVM2.AS.flash.text {
...
+
+    hasStyle(styleName: string): boolean {
+      return !!this._rules[styleName.toLowerCase()];
+    }
+
+    clear() {
+      this._rules = Object.create(null);
+    }
+
+    transform(formatObject: Style): flash.text.TextFormat {
```
tschneidereit: `formatObject` needs to be JS-ified before being used here.

tschneidereit:
So this is pretty cool! I like how simple it all is and how little code is involved.

PR #2011, opened 2015-01-17
Misc text fixes

tobytailor:


tobytailor:
Current reference image for firefox-textfield-align-ref-item1 is wrong. A more accurate result will be generated after merging.

PR #2012, opened 2015-01-18
Bug 1123051 - Improve compatibility of the main event loop's handling of Loader events

tschneidereit:


src/flash/display/AVM1Movie.ts
```diff
@@ -45,27 +45,22 @@ module Shumway.AVM2.AS.flash.display {
...
       this._setDirtyFlags(DisplayObjectFlags.DirtyChildren);
       this._invalidateFillAndLineBounds(true, true);
-      DisplayObject._advancableInstances.push(this);
+      display.DisplayObject._advancableInstances.push(this);
+      this._constructed = false;
     }
 
+    private _content: MovieClip;
+    private _constructed: boolean;
+
```
tobytailor: This should be renamed to not be confused with DisplayObjectFlags.Constructed which we set to indicate that the constructor was already called on this object. Or maybe completely getting rid of this extra property and introducing DisplayObjectFlags.FirstFrameReady or DisplayObjectFlags.FirstFrameConstructed instead?

src/flash/display/DisplayObjectContainer.ts
```diff
@@ -94,6 +94,7 @@ module Shumway.AVM2.AS.flash.display {
           child.dispatchEvent(events.Event.getInstance(events.Event.AVM1_INIT));
           child.dispatchEvent(events.Event.getInstance(events.Event.AVM1_CONSTRUCT));
           child._setFlags(DisplayObjectFlags.NeedsLoadEvent);
+          this._setFlags(DisplayObjectFlags.ContainsFrameScriptPendingChildren);
```
tobytailor: Shouldn't this be set conditionaly, only if HasFrameScriptPending is set on child?

src/flash/display/LoaderInfo.ts
```diff
@@ -72,24 +75,27 @@ module Shumway.AVM2.AS.flash.display {
...
-      this._frameRate = file.frameRate;
-      var bbox = file.bounds;
-      this._width = bbox.xMax - bbox.xMin;
-      this._height = bbox.yMax - bbox.yMin;
-      this._colorRGBA = file.backgroundColor;
-
-      if (!file.attributes || !file.attributes.doAbc) {
-        this._actionScriptVersion = ActionScriptVersion.ACTIONSCRIPT2;
+      if (file instanceof SWFFile) {
+        // TODO: remove these duplicated fields from LoaderInfo.
```
tobytailor: That + throwing "The loading object is not sufficiently loaded to provide this information" errors in their getters if this._file is not set.

src/flash/display/Loader.ts
```diff
@@ -83,77 +82,98 @@ module Shumway.AVM2.AS.flash.display {
...
-
     /**
-     * Handles the load status and dispatches progress events. This gets manually triggered in the
-     * event loop to ensure the correct order of operations.
+     * In each turn of the event loop, Loader events are processed in two batches:
+     * first INIT and COMPLETE events are dispatched for all active Loaders, then
+     * OPEN and PROGRESS.
+     *
+     * A slightly weird result of this is that INIT and COMPLETE are dispatched one
+     * turn *later* than the other events: INIT is only dispatched if the loadStatus
```
tobytailor: This is not necessarily related. It is true that INIT events (and therefore COMPLETE as well, since it depends on INIT being fired first) are dispatches AT LEAST one turn later, since INIT has to be fired AFTER the first frame is completely loaded and all its initial children are constructed (which, depending on download speed, can take longer than a single event loop turn). Again, a DisplayObjectFlags.FirstFrameReady or DisplayObjectFlags.FirstFrameConstructed flag could help here to decide whether to dispatch INIT.

It also looks like that processing loading events for the root loader takes one complete turn exclusively, meaning all loaders created on the very first frame are getting processed starting from frame 2. That really only is the case on the very first turn of the event loop. All loaders created in frames after that, are processed within the same turn.

test/swfs/as3-loader/events/loader-events.swf.trace
```diff
@@ -0,0 +1,36 @@
...
+child 1 loader: progress 8289
+child 2 loader: open 0
+child 2 loader: progress 0
+child 2 loader: progress 8289
+enterFrame 3
+added instance5
+construct instance2, parent: [object Loader] (TODO: parent should be null at this point, and the previous ADDED not dispatched)
+added instance2
+added instance6
+construct instance4, parent: [object Loader] (TODO: parent should be null at this point, and the previous ADDED not dispatched)
```
tobytailor: The parent being null here should only be true for root objects and not all timeline objects per se. Looks like Flash is adding a loader's content child AFTER its constructor got called. That explains why its parent is null and no ADDED events are fired at this point.

tobytailor:
Ok, I'm done with my review. Pretty happy to see this finally being implemented correctly.

src/flash/display/AVM1Movie.ts
tschneidereit: I think it could just use `this._content._hasFlags(DisplayObjectFlags.Constructed`, really. It's pre-existing and doesn't really matter, so I'm not all that motivated to change it in this PR.

src/flash/display/DisplayObjectContainer.ts
tschneidereit: You're right, except test should, I think, be `child._hasAnyFlags(DisplayObjectFlags.HasFrameScriptPending | DisplayObjectFlags.ContainsFrameScriptPendingChildren)`. I added that.

src/flash/display/Loader.ts
tschneidereit: > This is not necessarily related.

This only seems relevant if `COMPLETE` is dispatched in the same turn as the last `PROGRESS` event for loads that take longer than one turn of the event loop (i.e. where `INIT` and `COMPLETE` are dispatched in different turns.) I didn't test that and it doesn't seem too relevant, so I'd argue for keeping the current order of dispatching events, where we don't need to do any additional checks with newly introduced flags.

> It also looks like that processing loading events for the root loader takes one complete turn exclusively, meaning all loaders created on the very first frame are getting processed starting from frame 2. That really only is the case on the very first turn of the event loop. All loaders created in frames after that, are processed within the same turn.

If you have a test case for this, please send it my way. I don't fully understand the description in the abstract.

src/flash/display/LoaderInfo.ts
tschneidereit: Ah, good point about the exceptions. Will add those + a test.

test/swfs/as3-loader/events/loader-events.swf.trace
tschneidereit: Please provide test cases with comments like this. You clearly created a test case that shows this, so it doesn't make sense for me to try to figure out the details of it all over again.

src/flash/display/Loader.ts
tobytailor: In your test case, simple try to create the 2 Loader instances in the second frame of the root timeline instead of the first one. You will see that their `OPEN` and `PROGRESS` events will fire right after `EXIT_FRAME`, withing the same event loop turn, rather then one turn later.

PR #2028, opened 2015-01-25
Port most remaining traits from AS3 to TS

tschneidereit:
I left the net connection and stream stuff alone because @yurydelendik is working on those. Also, `SharedObject` looks very similar to those, so I wanted to hold off on that, too.

Also, tslint is choking on this. Hard. It spits out loads and loads of totally nonsensical errors and I have no idea why. @yurydelendik, maybe you do?


tschneidereit:
Ok, turns out tslint can't deal with `protected` yet. That actually seems fair, given that TypeScript 1.4 completely changed the way traversing over the AST works. What's not entirely fair is how utterly useless and spammy the results are.

PR #2040, opened 2015-01-31
Misc video player fixes

tobytailor:


extension/firefox/content/ShumwayStreamConverter.jsm
```diff
@@ -909,7 +911,11 @@ ShumwayStreamConverterBase.prototype = {
       throw new Error('Movie url is not specified');
     }
 
-    baseUrl = objectParams.base || pageUrl;
+    if (objectParams.base) {
+        baseUrl = new URL(objectParams.base, pageUrl).href;
```
yurydelendik: Can you use Services.io.newURI and [resolve()](http://mxr.mozilla.org/mozilla-central/source/netwerk/base/nsIURI.idl#178) method and then result's spec property? (Not sure we need to use Cu.importGlobalProperties(['URL']) for that)

PR #2041, opened 2015-01-31
Baseline JIT wireframe, WIP

mbebenita:


mbebenita:
It's barebones, but here it is.

I'm running it with `js build/ts/shell.js -c test/abcs~/*`

PR #2044, opened 2015-02-03
Sync with Bug 870553 whitelist changes.

yurydelendik:
See https://bugzilla.mozilla.org/show_bug.cgi?id=870553


yurydelendik:
`shumway.swf.whitelist` entry is a string in form of comma separated  `[@page_url object_url|@page_url|object_url]` entries,  e.g.

  `@http://example.org/* http://example.org/t.swf,http://www.mozilla.org/*.swf`

(enables t.swf on all example.org site pages and any swf from www.mozilla.org)

PR #2052, opened 2015-02-04
Replace XRegExp

tobytailor:


src/avm2/native.ts
```diff
@@ -1369,13 +1369,60 @@ module Shumway.AVM2.AS {
...
+        this.extended = false;
+        var captureNames = this._captureNames = [];
+        var p = pattern;
+        if (flags) {
+          var f = '';
+          for (var i = 0; i < flags.length; i++) {
+            var flag = flags[i];
+            if (flag === 's') {
+              // With this flag set, . can match the newline character.
+              p = p.replace(/\./g, '[\\s\\S]');
```
yurydelendik: what is regexp has `\\.`?

yurydelendik:
The input regular expression has to be fully parsed and reconstructed to EcmaScript one to make it work.

src/avm2/native.ts
```diff
@@ -1063,14 +1063,17 @@ module Shumway.AVM2.AS {
       if (re === (void 0) || re === null) {
         return null;
       } else {
-        if (re instanceof XRegExp && re.global) {
-          var matches = [], m;
-          while ((m = re.exec(this))) {
-            matches.push(m[0]);
+        if (re instanceof ASRegExp) {
```
tschneidereit: Because of the way we initialize AS classes, you have to use `AS.ASRegExp` here. Otherwise, it won't find the correct constructor. This will change in the future, but for now it is what it is.

tschneidereit: (Well, testing this it seems like the `instanceof` actually works. I don't know why it would, but great. Still, as a general rule always use at least the last part of the package path in situations like this.)

src/avm2/native.ts
```diff
@@ -1063,14 +1063,17 @@ module Shumway.AVM2.AS {
...
+                matches.push(m[0]);
+            }
+            return matches;
           }
-          return matches;
+          return re.match(<string><any>this);
         }
-        if (!(re instanceof XRegExp) && !(typeof re === 'string')) {
+        if (typeof re !== 'string') {
           re = String(re);
```
tschneidereit: Pre-existing, but this should be `asCoerceString`. And I would remove the `if` and just do the coercion.

src/avm2/native.ts
```diff
@@ -1369,71 +1372,127 @@ module Shumway.AVM2.AS {
...
+              // With this flag set, you can type spaces in the regular expression, which are
+              // ignored as part of the pattern.
+              p = p.replace(/\s+/g, '');
+              this.extended = true;
+            } else {
+              f += flag;
+            }
+          }
+        }
+        p = p.replace(/\((?!\?)|\(\?P<([\w$]+)>/g, function (match) {
```
tschneidereit: Something's wrong with this algorithm. In `test 1` of `test/avm2/pass/reg-0`, all group names are set to `?` instead of the proper group names.

src/avm2/native.ts
```diff
@@ -1369,71 +1372,127 @@ module Shumway.AVM2.AS {
...
     get native_extended(): boolean {
-      var self: any = this;
-      return self.extended;
+      return this.extended;
     }
 
     exec(s: string = ""): any {
-      var result = RegExp.prototype.exec.apply(this, arguments);
+      var result = this._jsRegExp.exec(s);
+
```
tschneidereit: Nit: lose the new-line.

src/avm2/native.ts
```diff
@@ -1369,71 +1372,127 @@ module Shumway.AVM2.AS {
...
+      var captureNames = this._captureNames;
+      for (var i = 0; i < captureNames.length; i++) {
+          var name = captureNames[i];
+          if (name !== null) {
+              result[name] = result[0] || '';
+          } else {
+              // Groups in AS3 are set to the empty string instead of undefined.
+              result[name] = '';
           }
```
tschneidereit: This `if .. else` condition is wrong. In the `else` block, all that can ever happen is that `name` is null, so it effectively boils down to `result[null] = ''`. I think you can just omit the `else` entirely, though, and things should just work.

src/avm2/native.ts
```diff
@@ -1369,71 +1372,127 @@ module Shumway.AVM2.AS {
...
-          if (result[k] === undefined) {
-            result[k] = "";
+
+      var captureNames = this._captureNames;
+      for (var i = 0; i < captureNames.length; i++) {
+          var name = captureNames[i];
+          if (name !== null) {
+              result[name] = result[0] || '';
+          } else {
+              // Groups in AS3 are set to the empty string instead of undefined.
```
tschneidereit: s/Groups in AS3/In AS3, unmatched groups/

src/avm2/native.ts
```diff
@@ -1369,71 +1372,127 @@ module Shumway.AVM2.AS {
...
+              result[name] = '';
           }
-        }
       }
+
       Shumway.AVM2.Runtime.publicizeProperties(result);
```
tschneidereit: I think it'd be better to create the public names when parsing the capture group names, so you don't have to do anything here: `publicizeProperties` uses `delete`, which isn't ever nice. When parsing names, you can just use `isNumeric` like so:
`publicName = isNumeric(name) ? name : '$Bg' + name;`

src/avm2/native.ts
```diff
@@ -1369,71 +1372,127 @@ module Shumway.AVM2.AS {
...
       return result;
     }
+
+    search(s: string): number {
+      return s.search(this._jsRegExp);
```
tschneidereit: Use `s.$Bgsearch` here, or `ASString.search.call(this._jsRegExp)`. Otherwise, you'll get the JS version of `String.prototype.search`.

src/avm2/native.ts
```diff
@@ -1369,71 +1372,127 @@ module Shumway.AVM2.AS {
...
+    }
+
+    match(s: string): string[] {
+      return s.match(this._jsRegExp);
```
tschneidereit: Same as for `search`.

tschneidereit:
Contrary to what I thought when we discussed this on IRC, I'm not sure there's anything fundamentally wrong with this approach. The wrong results might just be caused by the implementation bugs I pointed out (and perhaps ones I didn't see). Try what happens if you fix them and let's talk again, then.

src/avm2/native.ts
tobytailor: It is intended to call the native search here or else we end up in infinite recursions since this method itself gets called from ASString.search.

tobytailor:
This work merged into redux.

PR #2056, opened 2015-02-05
Changin default whitelist when the shumway.disabled=false

yurydelendik:
@tschneidereit is it what you had in mind for the preserving user choice when we update shumway.disabled?

### DIFF
--- a/extension/firefox/content/ShumwayBootstrapUtils.jsm
+++ b/extension/firefox/content/ShumwayBootstrapUtils.jsm
@@ -53,7 +53,9 @@ function getStringPref(pref, def) {
 }
 
 function log(str) {
-  dump(str + '\n');
+  var msg = 'ShumwayBootstrapUtils.jsm: ' + str;
+  Services.console.logStringMessage(msg);
+  dump(msg + '\n');
 }
 
 // Register/unregister a constructor as a factory.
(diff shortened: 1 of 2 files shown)


extension/firefox/content/ShumwayBootstrapUtils.jsm
```diff
@@ -89,6 +104,14 @@ var ShumwayBootstrapUtils = {
     if (registerOverlayPreview) {
       var ignoreCTP = getBoolPref(PREF_IGNORE_CTP, true);
       var whitelist = getStringPref(PREF_WHITELIST);
+      // Some platforms cannot support video playback, and our whitelist targets
+      // only video players atm. We need to disable Shumway for those platforms.
+      if (whitelist && !Services.prefs.prefHasUserValue(PREF_WHITELIST) &&
+          !allowedPlatformForMedia()) {
+        log('Default SWF whitelist is used on the unsupported platform -- ' +
```
tschneidereit: s/the/an/

extension/mozcentral/browser/extensions/shumway/content/ShumwayUtils.jsm
```diff
@@ -56,6 +58,19 @@ let ShumwayUtils = {
...
+  migratePreferences: function migratePreferences() {
+    // At one point we had shumway.disabled set to true by default,
+    // and we are trying to replace it with shumway.swf.whitelist:
+    // checking if the user already changed it before to reset
+    // the whitelist to '*'.
+    if (Services.prefs.prefHasUserValue(PREF_DISABLED) &&
+        !Services.prefs.prefHasUserValue(PREF_WHITELIST) &&
+        !getBoolPref(PREF_DISABLED, false)) {
+      // The user is already using Shumway -- enabling all web sites.
+      Services.prefs.setCharPref(PREF_WHITELIST, '*');
```
tschneidereit: I don't think we should do this. It'd mean that we can't ever get these users to the whitelist we want them to have. Instead, we should do something like
```js
if (this.enabled)
  this._ensureRegistered(getWhitelist());
```
Where `migratePreferences` replaced by a `getWhitelist` that just returns `'*'` for the users that have the `disabled` flag toggled, and the real whitelist from the config for everyone else.

yurydelendik: boolean preferences save prefHasUserValue flags for brief moment until prefrences are updated. If we will not record this somewhere else, we will loose this information.

On other hand, users can revert the whitelist by simple resetitng of the preference.

tschneidereit: For posterity's sake: we discussed this on IRC and I agree that, while it sucks, this is the most sensible way to deal with the issue.

PR #2063, opened 2015-02-07
Adds a way to trace call execution.

mbebenita:
This is very useful for debugging. To enable, set

```
var traceWriter = new IndentingWriter(false, dumpLine);
```

You will also need to make sure that the release flag is not set to true. At the moment we do this in utilities.ts and in viewerPlayer.js. Additionally, you will need to toggle the pref browser.dom.window.dump.enabled;true and naturally, run FF from the command line so you can get standard output.

### DIFF
--- a/src/avm2/runtime.ts
+++ b/src/avm2/runtime.ts
@@ -84,6 +84,8 @@ interface Function {
 }
 
 module Shumway.AVM2.Runtime {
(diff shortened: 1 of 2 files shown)


mbebenita:
Stale, and the functionality already exists in the recently landed redux branch.

PR #2064, opened 2015-02-07
Redux, WIP

mbebenita:
This PR is a set of changes that modify core VM structures in an attempt to lazify execution. The first commit, e7d6a74 reduces parsing time by nearly 4x. The idea is to get up and running as quickly as possible.


mbebenita:
@tschneidereit I'm opening this PR so you can keep track of the progress.

The parser uses a combination of lazy and eager evaluation. Part of the problem when parsing ABCs is that most data structures are of varying lengths. This makes skipping over them somewhat expensive. In general, if it's easy to skip over, I skip over and record the starting position so that I can evaluate it lazily at a later point. If it's not that easy to skip over, I'm usually better off just constructing the data structure in place.

Strings, Multinames, Namespaces, Namespace Sets are all lazy, the rest are a combination of eager and lazy. I'm pretty sure that the new setup also saves a bunch on memory.

The final parser API will deal with both: numeric IDs and lazily evaluated values. 

The next step will be to remove bytecode normalization and update the interpreter to deal directly with ABC byte streams.

yurydelendik:
Part of #2155

PR #2066, opened 2015-02-07
Fix performance regression introduced by 394e196

tschneidereit:


mbebenita:
Can you elaborate? I'm worried because we're doing the same thing in j2me.js

tschneidereit:
This regression that lasted about a day was caused by the change: http://arewefastyet.com/#machine=28&view=breakdown&suite=shumway
AWFY updates Shumway once per day, so it picked up this fix during the next update, eliminating the regression. And yes, it's certainly advisable to check the effect this has on j2me.js.

PR #2072, opened 2015-02-10
Remote debugger

yurydelendik:
Steps to test:
0. `grunt firefox`
1. Start `grunt exec:debug_server` (in addition to `grunt server`)
2. Open http://localhost:8000/examples/inspector/inspector.html and turn on inspector's "remoteEnabled" preference, reload the page (w/o `rfile` query parameter)
3. Install the extension, and create boolean pref in about:config `shumway.debug.enabled` = true
4. Open page with SWF in new instance of browser and in context menu select `Debug this SWF`.

What's happening behind scenes:
1. debug server starts HTTP server on port 8010, all requests will have session-id, from and to parameters
2. on start, inspector connects to debug server to listen for debugController/1/2;  and debug/<someUniqueId>/1;
3. on extension instantiation, the browser sends message to debugController/2/1 with message `{action: "getDebugger", swfUrl: url}`
4. the inspector, if debugController/1/2 gets url that matches `remoteSWF` options, it will return someUniqueId (from step 2), otherwise 0;
5.if extension gets value >= 2 (see step 2 and 4), it starts relaying all other messages over debug/1/<someUniqueId>

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -103,6 +103,9 @@ module.exports = function(grunt) {
                                    (grunt.option('rebuild') ? ' -r' : ''),
         cwd: 'utils/playerglobal-builder'
       },
+      debug_server: {
+        cmd: 'node examples/inspector/debug/server.js'
+      },
       gate: {
         cmd: '"utils/jsshell/js" build/ts/shell.js -x -g ' +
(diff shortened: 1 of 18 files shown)


yurydelendik:
See also https://github.com/mozilla/shumway/wiki/Remote-Debugging-of-Firefox-Extension

PR #2075, opened 2015-02-12
Misc yt fixes

yurydelendik:


yurydelendik:
Test url http://test.local:8000/examples/inspector/inspector.html?yt=4u-30V2Lq5I -- now stuck in iloop in ad content.

PR #2090, opened 2015-02-26
Refactoring of the ShumwayCom

yurydelendik:
* Removes FirefoxCom
* Moves Chrome privileged features from the stream converter
* Simplifies ShumwayCom

### DIFF
--- a/extension/firefox/chrome/ShumwayCom.jsm
+++ b/extension/firefox/chrome/ShumwayCom.jsm
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Mozilla Foundation
+ * Copyright 2015 Mozilla Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


extension/firefox/chrome/ShumwayCom.jsm
```diff
@@ -0,0 +1,701 @@
...
+    return shumwayComAdapter;
+  },
+
+  createActions: function (startupInfo, window, document) {
+    return new ShumwayChromeActions(startupInfo, window, document);
+  }
+};
+
+
+// All the priviledged actions.
```
tschneidereit: nit: "privilege".

extension/firefox/chrome/ShumwayCom.jsm
```diff
@@ -0,0 +1,701 @@
...
+    features: [],
+    errors: []
+  };
+
+  this.onLoadFileCallback = null;
+  this.onExternalCallback = null;
+}
+
+ShumwayChromeActions.prototype = {
+  invoke: function (name, args) {
```
tschneidereit: At least this function should have a comment explaining what it's for and why it's safe.

tschneidereit:
This is an extremely nice simplification! So, so much easier to understand and verify.

Land away! (But the two nits I mentioned should be addressed.)

PR #2103, opened 2015-03-06
Misc parser fixes

tobytailor:


src/flash/display/Sprite.ts
```diff
@@ -171,6 +171,7 @@ module Shumway.AVM2.AS.flash.display {
             var symbol: Shumway.Timeline.DisplaySymbol = null;
             if (hasCharacter) {
               symbol = <Shumway.Timeline.DisplaySymbol>loaderInfo.getSymbolById(placeObjectTag.symbolId);
+              // The Flash Player ignores references to not defined symbols here. So should we.
```
tschneidereit: Do we pick the symbol up on a later playback of the same frame if it became available in the meantime? The Flash Player does, AFAIR.

src/swf/SWFFile.ts
```diff
@@ -212,12 +210,18 @@ module Shumway.SWF {
         this._decompressor = Inflate.create(true);
         // Parts of the header are compressed. Get those out of the way before starting tag parsing.
         this._decompressor.onData = this.processFirstBatchOfDecompressedData.bind(this);
+        this._decompressor.onError = function (error) {
+          throw new Error(error);
```
tschneidereit: I'm pretty sure just throwing some error here isn't what we really want to do. No need to fix it now, but please add a TODO. We probably need to store the error on the FileLoader and call an error callback on the Loader, which can then do a `throwError` to content.

src/swf/SWFFile.ts
```diff
@@ -212,12 +210,18 @@ module Shumway.SWF {
...
+        }
         this._decompressor.push(initialBytes.subarray(8));
       } else if (isLzmaCompressed) {
         this.data.set(initialBytes.subarray(0, 8));
         this._uncompressedLoadedLength = 8;
         this._decompressor = new LzmaDecoder(true);
         this._decompressor.onData = this.processFirstBatchOfDecompressedData.bind(this);
+        this._decompressor.onError = function (error) {
+          throw new Error(error);
```
tschneidereit: Same here, obviously.

src/swf/parser/button.ts
```diff
@@ -31,22 +31,26 @@ module Shumway.SWF.Parser {
...
-      };
-      if (character.stateUp)
-        states.up.push(cmd);
-      if (character.stateOver)
-        states.over.push(cmd);
-      if (character.stateDown)
-        states.down.push(cmd);
-      if (character.stateHitTest)
-        states.hitTest.push(cmd);
+      // The Flash Player ignores references to not defined symbols here. So should we.
```
tschneidereit: nit: s/not defined/undefined/

Also, same question as above: do we pick the symbols up later? In this case, I don't at all know what the Flash Player does, though.

src/swf/parser/templates.ts
```diff
@@ -127,10 +127,10 @@ module Shumway.SWF.Parser {
     return readSb($bytes, $stream, size) / 65536;
   }
 
-  export function readString($bytes, $stream, length) {
+  export function readString($bytes, $stream, length?) {
```
tschneidereit: Given that this is hot code, I'd much prefer using `-1` for "not specified" and not making the `length` argument optional.

src/flash/display/Sprite.ts
tobytailor: Here we do, yes, since this code gets executed every time we navigate to this frame.

PR #2111, opened 2015-03-10
Bug 1141375 - Refactoring ShumwayCom to use Cu.cloneInto

yurydelendik:
- Defines separate set on functions for gfx and player
- Defines RtmpXHR/RtmpSocket interfaces
- Replaces function exports with cloneInto
- Replaces callback properties with setXXXXCallback listeners


yurydelendik:
https://bugzilla.mozilla.org/show_bug.cgi?id=1141375

PR #2113, opened 2015-03-11
Setup test harness and automate parse and playback tests

tobytailor:


tschneidereit:
There's a Flash test suite by Adobe called ATS (Adobe Test Suite, IINM). Let's not reuse the acronym for something entirely different - that's terribly confusing.

mbebenita:
What's the status on this? Is it ready to land?

PR #2115, opened 2015-03-11
Allows shuobject to be used crossdomain.

yurydelendik:
with some limitations (external interface and no access to internals)


yurydelendik:
Now web devs can write something like:

```
<!DOCTYPE html>
<html>
<head lang="en">
  <meta charset="UTF-8">
  <script src="https://cdn.rawgit.com/mozilla/shumway-dist/master/shuobject.js"></script>
  <script>
    // Create/embed SWF movie.
    shuobject.embedSWF('/shumway/demo.swf', 'mainSWF', 800, 512, '9,0,10');
  </script>
</head>
<body>
  <div id="mainSWF">Flash will be loaded here (it is similar to swfobject)</div>
</body>
</html>
```

to display SWF without installing shumway-dist repo

PR #2117, opened 2015-03-12
Use more efficient string concatenation in base64ArrayBuffer().

nnethercote:
This reduces the number of intermediate strings in the loop from
four to two. When loading 10 Flash ads, this reduces the number of
cumulative bytes allocated for intermediate strings from 7.6 MB to 3.8
MB.

### DIFF
--- a/src/base/utilities.ts
+++ b/src/base/utilities.ts
@@ -1058,7 +1058,8 @@ module Shumway {
         d = chunk & 63; // 63 = 2^6 - 1
 
         // Convert the raw binary segments to the appropriate ASCII encoding
-        base64 += encodings[a] + encodings[b] + encodings[c] + encodings[d];
+        base64 += concat4(encodings[a], encodings[b], encodings[c],
+                          encodings[d]);
       }
 
       // Deal with the remaining bytes and padding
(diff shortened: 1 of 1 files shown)


src/base/utilities.ts
```diff
@@ -1058,7 +1058,8 @@ module Shumway {
         d = chunk & 63; // 63 = 2^6 - 1
 
         // Convert the raw binary segments to the appropriate ASCII encoding
-        base64 += encodings[a] + encodings[b] + encodings[c] + encodings[d];
+        base64 +=
+            concat4(encodings[a], encodings[b], encodings[c], encodings[d]);
```
tschneidereit: Nit: we wrap arguments to the function call in cases like this. No need to change, will fix before merging.

tschneidereit:
Great, thanks. Will merge later today.

nnethercote:
I did the wrapping change.

nnethercote:
@tscheidereit: can you merge this? Thank you.

tschneidereit:
Yes, I can and will. However, I'm on vacation this week, so if @tobytailor
or @mbebenita don't beat me to it, I'll do it next Monday.

tschneidereit:
Thanks for the patch!

PR #2119, opened 2015-03-13
Avoid copying pixel data in updateFromDataBuffer().

nnethercote:
updateFromDataBuffer() can copy pixel data from an array buffer into an
ImageData object unnecessarily. In one Flash ad I saw an unnecessary
copy of 19.2 MB of pixel data this way.

This also required fixing the types for ImageData's constructor. Till
did this once before in #1969 but it looks like it was overwritten by a
tsc update.


nnethercote:
One thing I'm uncertain about with this changeset: it assumes that it is safe for updateFromDataBuffer() to take ownership of the `dataBuffer` argument. I'm not certain if that's true.

tschneidereit:
I discussed this with @mbebenita and we haven't been able to convince ourselves that it's safe to take this patch. The interface between the player and the GFX system is unfortunately not that easy to follow so I'm not 100% certain, but we probably already retrieve at least some of the buffers from `DataBuffer._arrayBufferPool`. Even if not, we might do so later have no really good way of making sure that we'll catch issues with it at that point. We can do so by changing all our code to use `'use strict'` and sealing the buffer when using it like in your patch. Even then, though, the patch will only really save on allocation rate, not absolute allocation, because either we already reuse the buffer right now - making the patch incorrect - or we don't, freeing it for GC. Overall, if you don't think there are really good reasons for taking it, I'm inclined to close the PR.

nnethercote:
I've analyzed ~100 ads now and I've only seen this pattern come up two or three times. So I'm happy to not do it.

PR #2121, opened 2015-03-16
Escape string

nnethercote:
[This ad](http://areweflashyet.com/swfs/2b1c3484cfd1fded9366fb6394ec3a736512144ac4a5650213eab80b9342b75fcd9c73835b9aa3d5e89fe6ead44bd9d5b75861e475640900c404d81f33788845.swf) causes `Shumway.AVM2.Compiler.AST.escapeString` to be called with a 1.6 MB XML string. This exposes the fact that `escapeString` is very inefficient, primarily because it does the escaping in two passes, creating two new strings. It does this in order to decide whether to use single or double quotes in the final string. Shumway doesn't need this flexibility, and so it can easily be reduced to a single pass.

### DIFF
--- a/src/avm2/compiler/c4/ast.ts
+++ b/src/avm2/compiler/c4/ast.ts
@@ -84,8 +84,7 @@ module Shumway.AVM2.Compiler.AST {
   var escapeStringCache = Object.create(null);
 
   export function escapeString(str: string) {
-    var result, i, len, ch, original = str;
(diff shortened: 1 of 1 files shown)


tschneidereit:
Great, thanks. We'll have to port this over to the redux branch, but that'll probably consist of just copying the whole functions to another file.

PR #2123, opened 2015-03-19
Bug 1089930 - intermittent issue in loader, "TypeError: f._addFrame is not a function"

tobytailor:


mbebenita:
Can you include a test case as well?

PR #2125, opened 2015-03-21
Fix clipping

tobytailor:
Our current clipping in the 2D backend doesn't work if groups, containing multiple shapes, are used as the clipping nodes. Instead of calling `clip()` for every shape, which leads to the final clipping path being the intersection of all shape paths, we have to combine them into a single `Path2D` object.


src/gfx/2d/2d.ts
```diff
@@ -197,6 +198,7 @@ module Shumway.GFX.Canvas2D {
 
     set (state: RenderState) {
       this.clip.set(state.clip);
+      this.clipPath = state.clipPath;
```
mbebenita: All other assignments in this function copy the state. Here you are aliasing the `clipPath`. Is that correct?

src/gfx/module.ts
```diff
@@ -606,11 +608,32 @@ module Shumway.GFX {
...
       // us a meaningful value. That we can only get if the platform gives us any means at all to
       // get that value. Gecko does so in the form of mozCurrentTransform, most engines don't.
-      if (!('currentTransform' in CanvasRenderingContext2D.prototype) &&
-        'mozCurrentTransform' in CanvasRenderingContext2D.prototype) {
-        Object.defineProperty(CanvasRenderingContext2D.prototype, 'currentTransform', {
-          get: mozPolyfillCurrentTransform
-        });
+      // For Chrome, at least return whatever transform was set using setTransform to ensure
+      // clipping works in our 2D backend.
+      if (!hasCurrentTransform) {
```
mbebenita: This will not work if any of the transformation functions like `translate`, or `scale`, `rotate` are called on the context. If you don't plan to support this at least override them with stubs that throw assertions.

src/gfx/module.ts
```diff
@@ -620,7 +643,7 @@ module Shumway.GFX {
     })();
   }
 
-  /**
+  /**`
```
mbebenita: Type-o.

src/gfx/renderables/renderables.ts
```diff
@@ -165,7 +165,8 @@ module Shumway.GFX {
      * Render source content in the specified |context|. If specified, the rectangular |cullBounds| can be used to cull parts of the shape
      * for better performance. If specified, |Region| indicates whether the shape's fills should be used as clip regions instead.
```
mbebenita: Comment is out of date.

src/gfx/2d/2d.ts
mbebenita: After discussing this with @tobytailor it turns out this is in fact correct. However, we should add a comment as to why that is.

tobytailor:
Merged manually.

PR #2131, opened 2015-03-25
Bug 1146593 - Checking for hang in runFrameScripts.

yurydelendik:
gotoAndPlay is really expensive right now and http://areweflashyet.com/swfs/1e902e33c40af1d1f6296bde21379600b35576a027cc888f78c5afe792db04e68b6e57671704aeee2814ee0e061608d386021a7985fb8e17a041e1658a39e3f7.swf has lots of objects calling it.

Detecting hang in runFrameScripts.

see https://bugzilla.mozilla.org/show_bug.cgi?id=1146593

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2131)

### DIFF
--- a/src/flash/options.ts
+++ b/src/flash/options.ts
@@ -37,4 +37,8 @@ module Shumway.AVM2.AS {
   export var flvOption = flashOptions.register (
     new Shumway.Options.Option(null, "FLV support.", "string", "unsupported", "Defines how to deal with FLV streams.")
   );
+
+  export var scriptsTimeout = flashOptions.register (
+    new Shumway.Options.Option("", "Scripts timeout", "number", 1000, "Specifies script execution timeout before hang is declared (in ms).")
+  );
 }
--- a/examples/inspector/js/inspectorDebugging.js
+++ b/examples/inspector/js/inspectorDebugging.js
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+Shumway.AVM2.AS.scriptsTimeout.value = 0;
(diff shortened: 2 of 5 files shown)


src/flash/options.ts
```diff
@@ -37,4 +37,8 @@ module Shumway.AVM2.AS {
   export var flvOption = flashOptions.register (
     new Shumway.Options.Option(null, "FLV support.", "string", "unsupported", "Defines how to deal with FLV streams.")
   );
+
+  export var scriptsTimeout = flashOptions.register (
+    new Shumway.Options.Option("", "Scripts timeout", "number", 1000, "Specifies script execution timeout before hang is declared (in ms).")
+  );
```
tobytailor: Isn't 1 second a little bit low? The SWF spec recommends something from 5 to 20 seconds. Also, we might want to take the ScriptTimeoutSeconds in ScriptLimits tags (if present) into account.

yurydelendik: I was thinking more like 500ms -- for ads it shall be enough. Once we move out of process we can increase it (and include ScriptTimeoutSeconds into account).

mbebenita:
@yurydelendik  What should we do with this PR?

yurydelendik:
@mbebenita The issue still exists, so I guess I'll just rebase -- we just have to agree on timeout value.

tobytailor:
@yury can you please rebase this PR if this is still something we wanna land? Or please close if not.

PR #2154, opened 2015-04-16
Rewrite of most of the AVM2 component, plus lots of other changes to adapt to said rewrite

tschneidereit:
Won't land right now, but getting close.


tschneidereit:
botio reftest

tschneidereit:
Landed in squashed form in #2155.

PR #2158, opened 2015-04-23
Implement getLineOffset.

mbebenita:


src/flash/text/TextField.ts
```diff
@@ -719,9 +719,22 @@ module Shumway.AVMX.AS.flash.text {
...
-      notImplemented("public flash.text.TextField::getLineOffset"); return;
+      var lines = this._textContent.plainText.split('\r');
+      if (lineIndex < 0 || lineIndex >= lines.length) {
+        this.sec.throwError('RangeError', Errors.ParamRangeError);
+      }
+      var offset = 0;
+      for (var i = 0; i < lineIndex; i++) {
+        offset += lines[i].length + 1; // Length + `\r`
+      }
+      // TODO:  I've tried modifying the width of the text field so that lines wrap, but this doesn't seem
```
tschneidereit: Eventually we should move to only adding TODOs with bug numbers, but probably not yet.

tschneidereit:
r=me. Not ideal, but it'll do for now.

PR #2159, opened 2015-04-23
Fix bug in calling Vector constructors as functions

tschneidereit:
@mbebenita, a quick look would be appreciated. Makes crypto.swf from the v8 tests run.


mbebenita:
Looks good.

PR #2160, opened 2015-04-23
Only rig timers and Math.random in a special `deterministic` mode in the shell

tschneidereit:
Because it interferes with the normal operation of benchmarking SWFs.

@mbebenita, this needs a close look, because I probably forgot places to guard with this flag.


mbebenita:
This look good to me.

PR #2162, opened 2015-04-24
Fix property lookup bug that breaks lots of Haxe-compiled content

tschneidereit:
This fixes [swf archive entry 1692](http://swf.codeazur.com.br/?class=&method=&width=&height=&tag=#9af86a33262baadce2ad23bb2bbb053e5e9624dcee4d77620094e435d5377114).

@mbebenita, this doesn't regress anything in our test suites (amazingly, it also doesn't make a single additional test pass), but I'd appreciate a sanity-check of my reasoning. In particular, I'm not at all sure if using `scope.parent` for class traits is the right thing or if some other scope should be used.

### DIFF
--- a/src/avm2/abc/lazy.ts
+++ b/src/avm2/abc/lazy.ts
@@ -1585,9 +1585,8 @@ module Shumway.AVMX {
      */
     public getNamespace(i: number): Namespace {
       if (i < 0 || i >= this._namespaceOffsets.length) {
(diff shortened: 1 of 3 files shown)


mbebenita:
> Though something else might still be going on here, for all I know.

It may be that the scope is pushed on the scope stack right away.

mbebenita:
Rest looks good.

PR #2163, opened 2015-04-24
Ensure that ASObject template classes are never instantiated

tschneidereit:
@yurydelendik, please especially check the closure compiler-related stuff. I have verified that all `release || foo` stuff gets nicely folded away in the cc bundles.


yurydelendik:
The constant folding works good.

PR #2164, opened 2015-04-24
arguments-callee

tschneidereit:
Add support for arguments.callee


tschneidereit:
@tobytailor, can you check if this covers your use case?
@mbebenita, can you do a sanity-check on the stuff in `_interpret`?

tobytailor:
Yes, this fixes my use case.

PR #2165, opened 2015-04-24
Bug 1139296 - broadcast RENDER event on correct DisplayObject instead of template class

tschneidereit:
@tobytailor, that should fix the bug.


tobytailor:
r+

PR #2172, opened 2015-04-25
Use rAF throttling to throttle frame execution.

mbebenita:


tschneidereit:
What happens if two banners coordinate their playback via `ExternalInterface` and one becomes invisible. IIUC, that'd cause the other to stop, too. What about sound playback? I think we need to be slightly more clever about this. One possibility is to reset the clock on `ExternalInterface` call-ins and sound/video events. (Note that stopping silent video would probably be just fine.)

mbebenita:
> What happens if two banners coordinate their playback via ExternalInterface and one becomes invisible. IIUC, that'd cause the other to stop, too.

We could throttle down instead of stopping it. However, that would cause two banners to get out of sync or cause both to be slow. I suspect that most of the time, synchronized banners are in the viewport at the same time, (we should measure this somehow) so we should take this into consideration.

> What about sound playback?

Can we ignore this for our immediate goal.


> I think we need to be slightly more clever about this. One possibility is to reset the clock on ExternalInterface call-ins and sound/video events.

Agreed. I'll keep investigating.

mbebenita:
I tried the `examples/raf/all.html` in Chrome and it looks like no Frame events are fired when the SWF is outside of the viewport. Wouldn't they have the same problem as this PR?

tschneidereit:
>> What happens if two banners coordinate their playback via ExternalInterface and one becomes invisible. IIUC, that'd cause the other to stop, too.
> 
> We could throttle down instead of stopping it. However, that would cause two banners to get out of sync or cause both to be slow. I suspect that most of the time, synchronized banners are in the viewport at the same time, (we should measure this somehow) so we should take this into consideration.

I don't think we can assume that. The classical scenario are a horizontal banner at the top and a vertical one at the side coordinating their animations. The top one can easily be scrolled out of view while the side one is visible.

>>     What about sound playback?
> 
> Can we ignore this for our immediate goal.

I suppose so.

tschneidereit:
> I tried the examples/raf/all.html in Chrome and it looks like no Frame events are fired when the SWF is outside of the viewport. Wouldn't they have the same problem as this PR?

Depends on how clever they are about it. This brings up an interesting question: does Swiffy convert banners that use `ExternalInterface` and do this kind of synchronization?

src/gfx/easelHost.ts
```diff
@@ -136,6 +137,36 @@ module Shumway.GFX {
...
+     */
+    private _addThrottleGovernor() {
+      var self = this;
+      var isThrottledDown = false;
+      var lastTick = performance.now();
+      // Poll for changes in the rAF frequency.
+      requestAnimationFrame(function tick() {
+        var thisTick = performance.now();
+        var elapsed = thisTick - lastTick;
+        if (elapsed < 500) {
```
yurydelendik: Add a constant here e.g. MAX_RAF_DURATION_FOR_VISIBLE_DOCUMENT.

src/gfx/easelHost.ts
```diff
@@ -136,6 +137,36 @@ module Shumway.GFX {
...
+          if (isThrottledDown) {
+            self._sendFocusEvent(Shumway.Remoting.FocusEventType.DocumentVisible);
+            isThrottledDown = false;
+          }
+        } else if (!isThrottledDown) {
+          // If the last tick was over 500 ms ago, we are probably outside of the viewport. Send a |DocumentHidden|
+          // event to the player to stop execution.
+          isThrottledDown = true;
```
yurydelendik: I think there shall be a secondary timeout to avoid noise from false document hidden state when CPU hangs for couple of seconds. Let's add MAX_RAF_TIMEOUT_FOR_HIDDEN_DOCUMENT == e.g. 30 seconds before we declare document is hidden.

yurydelendik:
It will be better if we will prolyfill that with an event we are trying to introduce at the platform level (if we have plans)

@mbebenita shall we also start player with "hidden" document state, if a page is opened in the inactive new tab, there will no be rAF events issued (so I think we will run a swf in "visible" state).

Make sure we don't turn on the governor during testing -- we shall add an option to turn this on/off.

tobytailor:
Do we still have intentions to land this?

PR #2174, opened 2015-04-25
Factor most state from _interpret into InterpreterFrame instance

tschneidereit:
Four reasons for doing this:
 - It's easier to follow.
 - It makes debugging easier because it makes it trivial to skip over the setup of all this state instead of having to single-step through it.
 - It's very slightly faster (~5% by my rough measurements).
 - It paves the way for extracting rarely used bytecodes into a new function to make the main interpreter loop much lighter-weight and better optimizable.

 But really, the first two are the motivation.

### DIFF
--- a/src/avm2/int.ts
+++ b/src/avm2/int.ts
@@ -108,60 +108,76 @@ module Shumway.AVMX {
     }
   }
 
-  function _interpret(self: Object, methodInfo: MethodInfo, savedScope: Scope, callArgs: any [],
-                      callee: AXFunction) {
-    var applicationDomain = methodInfo.abc.applicationDomain;
-    var sec = applicationDomain.sec;
+  class InterpreterFrame {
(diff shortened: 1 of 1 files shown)


tschneidereit:
@mbebenita, please take a look. If you don't like it at all, I'll can it. It does help me with debugging, though.

mbebenita:
> It paves the way for extracting rarely used bytecodes into a new function to make the main interpreter loop much lighter-weight and better optimizable.

Another advantage is that it would make it easier to reuse the same `_interpreter` JS frame, for multiple AS frames.

mbebenita:
r+, I'm fine with the approach.

tschneidereit:
> Ah, but having local and stack be the same number of chars used to be so nice. But, I'll live with it :)

But the grammar! Think of the poor children who might read our source code and be led on dark, dark path!

PR #2179, opened 2015-04-26
Optimize namespace handling

tschneidereit:
By interning all namespaces and reducing the complexity of trait lookups. Builds on #2176 so can't land before that.


tschneidereit:
Forgot: r? @mbebenita.

PR #2180, opened 2015-04-26
Log startup timings

tschneidereit:
Logs the duration of two initialization phases: from startup triggering to player start and from player start to entering of the main loop.

@yurydelendik, I couldn't find a better way of funneling the startup timestamp through. If you see one, please tell me. Otherwise: r?

### DIFF
--- a/examples/inspector/inspector.html
+++ b/examples/inspector/inspector.html
@@ -83,8 +83,9 @@ limitations under the License.
       </div>
     </div>
     <script>
-      var requiredConsoleFunctions = ["profile", "profileEnd", "markTimeline",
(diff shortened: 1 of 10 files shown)


yurydelendik:
>  I couldn't find a better way of funneling the startup timestamp through. 

Could you add this time to the "flashParams" (it's a bad name -- we probably need to change that to "playerStartupParameters" or something) for the extension?

tschneidereit:
> Could you add this time to the "flashParams" (it's a bad name -- we probably need to change that to "playerStartupParameters" or something) for the extension?

Good call, done.

yurydelendik:
r+

PR #2183, opened 2015-04-26
Some AVM1 fixes

tschneidereit:
Makes SWF 123 from the archive show more stuff. Still not fully working, but progress.

r? @yurydelendik.

### DIFF
--- a/src/avm1/lib/AVM1MovieClip.ts
+++ b/src/avm1/lib/AVM1MovieClip.ts
@@ -75,7 +75,7 @@ module Shumway.AVM1.Lib {
 
     _lookupChildByName(name: string): AVM1Object {
       name = alCoerceString(this.context, name);
-      var lookupOptions = flash.display.LookupChildOptions.INCLUDE_NOT_INITIALIZED;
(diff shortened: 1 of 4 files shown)


src/avm1/lib/AVM1MovieClip.ts
```diff
@@ -343,7 +343,7 @@ module Shumway.AVM1.Lib {
       var nativeObject = this.as3Object;
       var maxDepth = 0;
       for (var i = 0, numChildren = nativeObject.numChildren; i < numChildren; i++) {
-        var child = nativeObject._lookupChildByIndex(i);
+        var child = nativeObject._children[i];
```
yurydelendik: I'm okay with that, but I would rather add a LookupChildOptions parameter to the _lookupChildByIndex (see _lookupChildByName)

tschneidereit: > I'm okay with that, but I would rather add a LookupChildOptions parameter to the _lookupChildByIndex (see _lookupChildByName)

Ok, changed in the new version. Again, r?

yurydelendik:
r+

PR #2196, opened 2015-04-29
Emit warnings only once, if not overridden by setting Shumway.loggingOptions.omitRepeatedWarnings to false

tschneidereit:
r? @yurydelendik


tschneidereit:
@yurydelendik, this implements what we talked about yesterday. Additionally, there's now a function `Shumway.Debug.warnCount()` that you can use to dump a list of encountered warnings and their frequency. Quite useful in my short experimentations.

Excellent example of where this is useful: [SWF 1575](http://swf.codeazur.com.br/#e3f8f4e110b61add10b448e65334c2c9d009c84e206c53811c1051aff289561c).

yurydelendik:
r+

PR #2205, opened 2015-04-30
Adds XMLDocument classes and parser.

yurydelendik:


tschneidereit:
r=me with feedback addressed.

PR #2210, opened 2015-04-30
Implement TextField#html

tschneidereit:
At least a mostly correct version of it.

r? @tobytailor.


tschneidereit:
Now with the test I meant to add.

tobytailor:
r+ but reminds me that we have to convert text runs back to html after parsing.

PR #2213, opened 2015-05-01
WIP: Implement BitmapData#getColorBoundsRect

tschneidereit:
And required fixes to BitmapData. Very WIP, much broken.

There is so much inconsistency in BitmapData, particularly around handling of color formats. I think we need to clean all that up, and sooner rather than later, to be able to say, with a straight face, that we support any methods on BitmapData at all.

In particular, various methods just assume the data to be in pARGB format, while others think it's uRGBA. With some uncertainty about byte order sprinkled in. The comment on `_ensureBitmapData` tells a story about converting things to uRGBA, but the implementation converts to pARGB.

Fun.

### DIFF
--- a/src/base/utilities.ts
+++ b/src/base/utilities.ts
@@ -3217,6 +3217,30 @@ module Shumway {
       return Math.max(0, Math.min(255, value));
     }
 
+    export function uARGBToImagePixel(value: number, imageType: ImageType) {
+      switch (imageType) {
+        case ImageType.PremultipliedAlphaARGB:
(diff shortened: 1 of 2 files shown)


tschneidereit:
Oh and also, `getColorBoundsRect` absolutely requires some unpremultiplied format. Which one doesn't matter much, though.

tobytailor:
@tschneidereit can you please resolve the mergin conflicts if this is still something we want to land. I can do the review then. If not, please close this PR.

PR #2219, opened 2015-05-03
Removes mongodb from devDependencies.

yurydelendik:
To avoid failures like https://travis-ci.org/mozilla/shumway/builds/61031448


yurydelendik:
(Needs only for setup of db for [Testing using ATS files](https://github.com/mozilla/shumway/wiki/Testing-using-ATS-files) )

PR #2222, opened 2015-05-03
Comment out some dead code

tschneidereit:
@mbebenita, you introduced this in ace7504064ec35d532078b2f0783a038b6318bf9. Is this something that should be changed, or can we just remove the now-disabled code?


yurydelendik:
Addressed by #2255

PR #2226, opened 2015-05-04
Guard during frame parsing when soundstreanhead is not supported.

yurydelendik:
Fixes this._sounds is null errors during parsing.

Needs ADPCM support for in-SWF stream sounds.


yurydelendik:
Some external botio configuration affected reference images (?)

PR #2232, opened 2015-05-06
Fix some Makefile typos

cpeterso:
s/Coping/Copying/ and change "Copy" and "Package" to use the same tense as the other Makefile comments.


yurydelendik:
Thanks

PR #2238, opened 2015-05-08
Removes TestEaselHost and TestPlayer classes.

yurydelendik:
... and adds an iframe for gfx classes. also, fixed swfm recorder and profiler.


tschneidereit:
AFAICT, this completely removes the ability to use the inspector without the iframe player. If that's the case: please undo it. It's still the case that debugging is more annoying across multiple iframes.

If there's a fundamental reason why undoing it is not possible, then please explain this reason and discuss options with other people in the team.

yurydelendik:
Please explain "debugging is more annoying across multiple iframes". I personally don't feel a difference between debugging in iframes vs without them.

Reason for this change is to reduce amount of configuration we have to work with (I discussed this with @mbebenita). Also reduce amount of regressions, e.g. in the past we had issues with text. Testing is happening in iframes mode as well.

This is first step for security work for sandboxing gfx iframe and it is blocked by this patch.

If there are major issue with debugging stuff, we shall file bugs for devtools. As last resort, I can create a special inspector without frames, similar to inspector-bundles.

tschneidereit:
Thanks for the explanation, it does make sense. I'll dig into what the exact differences are that make it still more annoying for me once I'm back from vacation.

Please make sure to document changes like this, including the reason, and ask for review. We should all do that for every non-trivial change going forward.

PR #2252, opened 2015-05-14
Fix AVM1 action block execution order.

tobytailor:


src/flash/display/MovieClip.ts
```diff
@@ -145,16 +171,36 @@ module Shumway.AVMX.AS.flash.display {
...
+        if (movieClipClass.frameNavigationModel === FrameNavigationModel.SWF1) {
+          // AVM1 action blocks must be executed exactly in the same order they appear in the SWF
+          // file. We keep track of the position for each action block at parsing time and carry that
+          // information on to their wrapping functions (via the precedence property). Here we queue
+          // up all such functions so we can sort and execute them after this loop.
+          if (instance._hasFlags(DisplayObjectFlags.NeedsLoadEvent)) {
+            instance._removeFlags(DisplayObjectFlags.NeedsLoadEvent);
+            release || assert(instance._symbol);
+            var handler = (<AVM1.Lib.AVM1SymbolBase>
+              AVM1.Lib.getAVM1Object(instance, instance._symbol.avm1Context))._onLoadHandler;
```
yurydelendik: _onLoadHandler looks like a hack, can we have something like 

```
var handler = function (instance) {
  instance.dispatchEvent(eventClass.getInstance(events.Event.AVM1_LOAD));
}.bind(null, instance);
```
?

src/flash/display/MovieClip.ts
```diff
@@ -145,16 +171,36 @@ module Shumway.AVMX.AS.flash.display {
...
+          for (var j = 0; j < frameScripts.length; j++) {
+            var script = frameScripts[j];
+            script.context = instance;
+            unsortedScripts.push(script);
+          }
           continue;
         }
-
-        instance._allowFrameNavigation = movieClipClass.frameNavigationModel === FrameNavigationModel.SWF1;
+
```
yurydelendik: nit: unneeded change

tobytailor: More an unneeded assignment. `_allowFrameNavigation` is `true` by default and this line never reached if `movieClipClass.frameNavigationModel === FrameNavigationModel.SWF1`.

src/flash/avm1.d.ts
```diff
@@ -34,7 +34,13 @@ declare module Shumway.AVM1 {
     function getAVM1Object(as3Object:flash.display.DisplayObject, context: AVM1Context): AVM1Object;
     function initializeAVM1Object(as3Object: flash.display.DisplayObject, context: AVM1Context,
                                   placeObjectTag: Shumway.SWF.PlaceObjectTag): void;
-    class AVM1MovieClip extends AVM1Object {
+    class AVM1 extends AVM1Object {
```
yurydelendik: unused class

src/flash/display/MovieClip.ts
```diff
@@ -114,6 +114,32 @@ module Shumway.AVMX.AS.flash.display {
     }
   }
 
+  export interface FrameScript {
+    (any?): any;
+    precedence?: number;
+    context?: MovieClip;
```
yurydelendik: Can we rename `context` here to something else, e.g. `symbol`?

src/flash/display/MovieClip.ts
```diff
@@ -114,6 +114,32 @@ module Shumway.AVMX.AS.flash.display {
     }
   }
 
+  export interface FrameScript {
+    (any?): any;
```
yurydelendik: Not sure what this notation mean, can we specify field names here?

yurydelendik:
I think it's good to go when _onLoadHandler logic is replaced by dispatching of avm1 'load' event (and comments above addressed)

src/flash/display/MovieClip.ts
tobytailor: I like context more. Any other recommendations? We already use the term symbol for objects to initialize instances from, not the instances itself.

src/flash/display/MovieClip.ts
tobytailor: This is simply the TypeScript way do define objects implementing FrameScript to be callable (a function).

src/flash/avm1.d.ts
yurydelendik: "class AVM1" is added by this commit, am I missing something?

src/flash/display/MovieClip.ts
tobytailor: We have to carry the precedence somehow, which is set as a property on the function `_onLoadHandler` (which implements the `FrameScript` interface) refers to. Retrieving that information from event handlers would result in a way uglier hack.

src/flash/avm1.d.ts
tobytailor: Mh, seems like you're right. Sorry. Have no idea how that happened...

src/flash/display/MovieClip.ts
yurydelendik: Okay, so I didn't fully understand how precedence is flowing. In this case we shall definitely remove _onLoadHandler and handling of precedence from AVM1 logic -- virtual machine shall not know about intricacies of timeline things. We can save precedence at placeobject (I don't think actionblock precedence matters in this case, worst case we need to get it from the placeobject events from actionsblock), and that will make much more sense.

I think, the _onLoadHandler hack will prevent us from handling onLoad events (http://help.adobe.com/en_US/AS2LCR/Flash_10.0/help.html?content=Part2_AS2_LangRef_1.html) that are bound dynamically.

yurydelendik:
looks good, now with swfdec tests

/botio test

yurydelendik:
thank you

PR #2255, opened 2015-05-15
Fixes "Flash" stroke scaling.

yurydelendik:


yurydelendik:
(We need to look into incorrect acid-scale circle/ellipse stroke later)

PR #2260, opened 2015-05-20
Remove jpgjs dependency, apply jpeg3 alpha values in the gfx backend.

tobytailor:


yurydelendik:
there is also 'src/swf/jpeg.ts' file.

mbebenita:
This patch needs a test case.

PR #2261, opened 2015-05-21
Refactors AVM1 object initialization and DoInitAction.

yurydelendik:
It's moving AVM1 script binding logic into _constructChildren(). The AVM1 specific methods were moved from AVM1MovieClip has to be moved (currently to the MovieClip) since during binding, and even before scripts are executed, AVM1 objects shall not exist.


yurydelendik:
bad solution

PR #2265, opened 2015-05-29
Fix file naming in the player module

tschneidereit:
And this is why all changes should go through review.


tschneidereit:
r=me, straight-forward renaming of files.

PR #2225, opened 2015-05-04
Implement LocalConnection

tschneidereit:
Not finished, contains only the content-facing parts. Those should be pretty much done, though, modulo bug fixes.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2225)

PR #2266, opened 2015-06-01
Bug 1068924 - Implement LocalConnection

tschneidereit:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2266)


tschneidereit:
This is the same as #2225, will address the latest review comments from that next.

tschneidereit:
Replying to @yurydelendik's comment from #2225:
> replace all LocalConnection members with getLocalConnectionManager which will return an object (as example see createSpecialStorage above) with these methods.

I don't think the situations are the same. Effectively the `localConnectionService` is what you want to have returned with `getLocalConnectionManager`: it contains the implementation of all the methods related to LocalConnections. However, I think it's best to separate the argument sanitizing from the actual implementation, so `ShumwayCom` itself contains wrappers for all methods of the service that do the sanitizing before calling the actual service methods.

I did fix a few issues with both the methods and the wrappers in an upcoming patch, though.

yurydelendik:
> However, I think it's best to separate the argument sanitizing from the actual implementation, so ShumwayCom itself contains wrappers for all methods of the service that do the sanitizing before calling the actual service methods.

That for encapsulation of entire LocalConnection functionality in one place (giving that some kind of namespace). From my point of view it's no different than SpecialInflate, RTMP, etc. Also for documentation sake too, e.g https://github.com/mozilla/shumway/wiki/Outer-Sandbox-Security -- somebody has to update that.

mbebenita:
> somebody has to update that

We should file follow-up bugs for documentation updates.

tschneidereit:
> We should file follow-up bugs for documentation updates.

I'll update the documentation before merging this. We shouldn't land changes like this without updating the docs.

tschneidereit:
@yurydelendik, ok I think this is ready for another review round.

yurydelendik:
I'm still not really comfortable with prefixing using qualifyLocalConnectionName on client side and trusting senderUrl in the sendLocalConnectionMessage in the chrome code for verifying origin. Because of that, could you add a kill switch on the chrome side in the prefs to disable any connection?

Once the comments addressed and tests pass -- it will be fine to land.

---

Review status: all files reviewed, 9 unresolved discussions, some commit checks failed.
Reviewed files:
- examples/inspector/js/inspectorPlayer.js @ r1
- extension/firefox/chrome/LocalConnection.jsm @ r3
- extension/firefox/chrome/ShumwayCom.jsm @ r3
- extension/firefox/content/web/viewerPlayer.js @ r1
- src/avm2/errors.ts @ r1
- src/avm2/natives/byteArray.ts @ r1
- src/base/external.ts @ r3
- src/base/utilities.ts @ r3
- src/flash/display/Loader.ts @ r2
- src/flash/events/AsyncErrorEvent.as @ r1
- src/flash/events/AsyncErrorEvent.ts @ r1
- src/flash/events/StatusEvent.as @ r1
- src/flash/events/StatusEvent.ts @ r1
- src/flash/link.ts @ r1
- src/flash/module.ts @ r1
- src/flash/net/LocalConnection.as @ r1
- src/flash/net/LocalConnection.ts @ r1
- src/flash/references.ts @ r2
- src/player/external.ts @ r3
- src/player/player.ts @ r2
- src/player/references.ts @ r1
- src/shell/domstubs.ts @ r1
- src/shell/playerservices.ts @ r1
- src/shell/shell.ts @ r1
- test/harness/slavePlayer.js @ r1
- test/swfs/localconnection/lc-receive.fla @ r1
- test/swfs/localconnection/lc-receive.swf @ r1
- test/swfs/localconnection/lc-send.fla @ r1
- test/swfs/localconnection/lc-send.swf @ r1
- test/swfs/localconnection/lc-send.swf.trace @ r1
- test/swfs/localconnection/LCReceive.as @ r1
- test/swfs/localconnection/LCSend.as @ r1
- test/test_manifest_trace.json @ r1
- test/test_mock.baseline @ r1

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 60 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvjbUgAITl8T5aOeJ_)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L60)):</sup>
add information if connection was created from secured location, e.g `secure: parsedURL.schemeIs('https')`

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 107 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqveW4XsMQyLSeQq5d5)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L107)):</sup>
This comment provided not a proper verification of "HTTP origin calls to HTTPS origins" -- we can remove that.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 108 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-Jqvh1MlRffwiSJWkJGx)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L108)):</sup>
you don't need to special case `connectionName[0] === '_'` here, allowDomain rules is applied to those the same way.

if connection.secure, add parsedURL.schemeIs('https') check in addition to the domain one.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 109 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-Jqvkj8aMG03ooFn7q-D)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L109)):</sup>
is allowedDomains undefined here?

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 157 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvkINT3ghHT-VSTXGX)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L157)):</sup>
replace that with logic equal to:
- if connection.secure, check only connection.allowedDomains
- otherwise, check both connection.allowedDomains and .allowedInsecureDomains

---

<sup>**[src/flash/net/LocalConnection.ts, line 41 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvY7prKkYwA-3rydAV)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/flash/net/LocalConnection.ts#L41)):</sup>
The URL (or at least origin) must be passed during connection construction (I suggest secure property in the chrome jsm code and populate allowDomain entry there) to ensuring a proper validation/enforcing of the "from non-HTTPS origins to send LocalConnection calls to files from HTTPS origins" rule.

---

<sup>**[src/flash/net/LocalConnection.ts, line 179 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-Jqvb0wuAHCz2lWGPko6)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/flash/net/LocalConnection.ts#L179)):</sup>
Per spec the LocalConnection class can be extended to contain methods on itself (in addition to the `client` property object). I don't think forbiddenNames check shall not apply for `client` object, but it certainly makes sense for `this`.

---

<sup>**[src/flash/net/LocalConnection.ts, line 230 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvgCy8_TGwt57W5lJQ)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/flash/net/LocalConnection.ts#L230)):</sup>
Why not use `new URL(this._url).hostname` here?

---

<sup>**[src/player/external.ts, line 333 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvhqmBkFrozJI_RYsk)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/player/external.ts#L333)):</sup>
assert in the qualifyLocalConnectionName for ':' will fail when domain-prefixed connectName is used

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

yurydelendik:
Eh, I provided wrong checks with allowInsecureDomains, correction:
- if connection.secure and url-to-check is not-secure, check allowInsecureDomains
- otherwise, allowDomains

---

Review status: all files reviewed, 9 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

yurydelendik:
Review status: all files reviewed, 9 unresolved discussions, some commit checks failed.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 157 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvkINT3ghHT-VSTXGX)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L157)):</sup>
Correction:
- if `connection.secure and !secure`, check connection.allowedInsecureDomains
- otherwise, check connection.allowedDomains 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

tschneidereit:
Thanks for the review.

I changed the PR quite a bit so would appreciate a re-review. Important changes:
- Added logging, behind a pref
- Added TODOs for telemetry so we don't forget that
- Added more asserts
- Changed all URL verification to rely on `environment.swfUrl` in chrome code rather than on whatever content passes in
- Fixed `allowDomain` checks to follow correct semantics, i.e. what you wrote above
- Added a check to `closeLocalConnection` to ignore attempts to close a connection from another Shumway instance

I re-enabled the whitelisting of domains: I realized that we can actually trust that part completely, as content is free to enable whichever domains it wants. I.e., if the receiver whitelisted a domain, then we can validly use that information. It's only the sender's domain that we can't trust as passed by player code.

What we need to do eventually, I think, is disallow LocalConnection messages from player instances that contain SWFs from multiple URLs. Or, check all of the URLs against the receiver's `allowDomain` lists, rather. We currently, AFAICT, only have the root SWF's URL readily available as `environment.swfUrl`, but it seems like we should be able to just add all loaded SWF's URLs to `environment`. I'll do that as a follow-up.

I did not add a kill-switch, because I don't think just disabling the chrome parts makes much sense. If we need to disable this, I'll do it manually.

Also, this all needs extension tests once we have the infrastructure for those in place.

---

Review status: 24 of 34 files reviewed, 7 unresolved discussions, some commit checks pending.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 108 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-Jqvh1MlRffwiSJWkJGx)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L108)):</sup>
Done.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 109 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-Jqvkj8aMG03ooFn7q-D)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L109)):</sup>
Yes, it was, thanks. Fixed.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 157 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvkINT3ghHT-VSTXGX)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/extension/firefox/chrome/LocalConnection.jsm#L157)):</sup>
Done.

---

<sup>**[src/flash/net/LocalConnection.ts, line 41 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvY7prKkYwA-3rydAV)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/flash/net/LocalConnection.ts#L41)):</sup>
I implemented this by, in chrome code, essentially ignoring what content passes in here. I didn't rip that out entirely because we'd have to re-add exactly the same thing and I like having it very explicit what's happening.

---

<sup>**[src/flash/net/LocalConnection.ts, line 179 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-Jqvb0wuAHCz2lWGPko6)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/flash/net/LocalConnection.ts#L179)):</sup>
The `_client` property is initialized to `this`, so the logic fully works. The forbidden method names are always forbidden, regardless of what the client is. They really shouldn't even reach this point, as the message won't be sent, so this is only relevant for misbehaving senders. Hence the comment below.

---

<sup>**[src/flash/net/LocalConnection.ts, line 230 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvgCy8_TGwt57W5lJQ)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/flash/net/LocalConnection.ts#L230)):</sup>
This was pre-existing code that I just moved here. Still, fixed to use `URL`.

---

<sup>**[src/player/external.ts, line 333 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JqvhqmBkFrozJI_RYsk)** ([raw file](https://github.com/mozilla/shumway/blob/e6481641f60eba853a948350ba954e692c2dba43/src/player/external.ts#L333)):</sup>
Thanks, fixed.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

yurydelendik:
Looks good to me with comments addressed.

---

Review status: all files reviewed, 5 unresolved discussions, all commit checks successful.
Reviewed files:
- examples/inspector/js/inspectorPlayer.js @ r4
- extension/firefox/chrome/LocalConnection.jsm @ r4
- extension/firefox/chrome/ShumwayCom.jsm @ r4
- extension/firefox/content/web/viewerPlayer.js @ r4
- src/base/external.ts @ r4
- src/base/utilities.ts @ r4
- src/flash/net/LocalConnection.ts @ r4
- src/player/external.ts @ r4
- test/harness/slavePlayer.js @ r4
- test/test_manifest_trace.json @ r4

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 35 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcHFQtmthbj-JCNhJG)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L35)):</sup>
There are several `typeof connectionName !== 'string' ||  connectionName[0] !== '_' && connectionName.split(':').length !== 2` checks in this file: extract that into function (e.g. `isConnectionNameVaild`). My recommendation to use `indexOf`s there instead of `split`.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 146 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcIHOI-UxxRWDzFEiV)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L146)):</sup>
Shall be `(parsedURL.protocol === 'https:' && !senderIsSecure)`

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 168 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcIeAoBA_PMdoVNCCU)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L168)):</sup>
nit: bracket `{` belongs to previous line (here and below)

---

<sup>**[src/flash/net/LocalConnection.ts, line 41 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcFWV0V7Ieb6hy43lg)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/src/flash/net/LocalConnection.ts#L41)):</sup>
`new (<any>URL)(...)` will be better transformed into JavaScript

---

<sup>**[src/player/external.ts, line 272 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcGYEVpeU-JTt_yA1F)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/src/player/external.ts#L272)):</sup>
let's rename `assertValidity` to `allowPrefix` (or `denyPrefix`?) -- we need to verify validity on the connectionName.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

tschneidereit:
Review status: 31 of 34 files reviewed, 5 unresolved discussions, some commit checks pending.
Reviewed files:
- extension/firefox/chrome/LocalConnection.jsm @ r4

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 35 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcHFQtmthbj-JCNhJG)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L35)):</sup>
Done. I kept the split, though, because `indexOf` would require a loop or an additional `lastIndexOf`: we want to make sure that there's exactly one `:` in the name. (If speed were important here, I wouldn't use `split`, of course.)

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 146 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcIHOI-UxxRWDzFEiV)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L146)):</sup>
No, that would be a less strict check: if `parsedURL.protocol` is `http`, the second part wouldn't be checked anymore. Here, we want to check if the parsed URL's information is identical to the client-provided. The check is way too subtle as-is, though, so I changed it to extract `parsedURLIsSecure`, making the check a straight-forward boolean comparison.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 168 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcIeAoBA_PMdoVNCCU)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L168)):</sup>
This is taken from SpiderMonkey's style which cuddles braces except for multiline statements, which get the brace on the next line. I changed it because we never stipulated that in our style guide, though.

---

<sup>**[src/flash/net/LocalConnection.ts, line 41 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcFWV0V7Ieb6hy43lg)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/src/flash/net/LocalConnection.ts#L41)):</sup>
D'oh, of course. Thanks.

---

<sup>**[src/player/external.ts, line 272 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcGYEVpeU-JTt_yA1F)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/src/player/external.ts#L272)):</sup>
I renamed it to `assertNoPrefix`: I want the `assert` in the name because this only affects asserts and doesn't have any effect in opt builds.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

yurydelendik:
Review status: all files reviewed, 1 unresolved discussion, all commit checks successful.
Reviewed files:
- extension/firefox/chrome/LocalConnection.jsm @ r5
- src/flash/net/LocalConnection.ts @ r5
- src/player/external.ts @ r5

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 35 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcHFQtmthbj-JCNhJG)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L35)):</sup>
loop is not needed, second indexOf check is okay.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

tschneidereit:
Review status: all files reviewed, 1 unresolved discussion, all commit checks successful.

---

<sup>**[extension/firefox/chrome/LocalConnection.jsm, line 35 \[r4\]](https://reviewable.io:443/reviews/mozilla/shumway/2266#-JrcHFQtmthbj-JCNhJG)** ([raw file](https://github.com/mozilla/shumway/blob/d4c577f7da9b7761ecfca72ef1455da122039f6d/extension/firefox/chrome/LocalConnection.jsm#L35)):</sup>
We discussed this on IRC, and I got yury to say "okay" :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2266)

PR #2267, opened 2015-06-01
WIP: Playing FLVs using VP6Player

yurydelendik:
Creates simple SWF that plays FLV files (VP6 and non-VP6). The flvOption setting now has addition options: `flash` and `alwaysflash`. The latter affects even non-FLV files -- mostly for testing.

Problem with the extension that FP runs SWF from 'file:///' location however player is hosted at 'resource://shumway/viewer.gfx.html' -- this does not allow to call anything from SWF: SecurityError: Error #2060: Security sandbox violation: ExternalInterface caller file:///Users/{user}/Library/Application Support/Firefox/Profiles/default/extensions/shumway@research.mozilla.org/content/vp6player.swf cannot access resource://shumway/web/viewer.gfx.html. An SWF path in particular user's profile above must be added in trusted location in FP configuration (Mac OSX: System Preferences -> Flash Player -> Advanced -> Trusted Location Settings...)

TODOs:
  - [ ] Transparency
  - [ ] Properly fail if the extension has no proper security configuration (see above)
  - [ ] appendBytes handling (or do it later?)


yurydelendik:
Related bug https://bugzilla.mozilla.org/show_bug.cgi?id=1137860

Known blockers:
- [ ] Object tag is not visible on Windows https://bugzilla.mozilla.org/show_bug.cgi?id=1169455
- [ ] Object tag location hosted at resource:// (see permissions issue above) https://bugzilla.mozilla.org/show_bug.cgi?id=1170729

tobytailor:
Is it worth fixing the remaining TODOs to land this?

PR #2270, opened 2015-06-03
Adds HTTPS server options to the internal web server.

yurydelendik:
Run `make -C test/ssl/ generate` to generate self-signed certificate.


yurydelendik:
Trivial

PR #2274, opened 2015-06-05
Bug 1171898 - Box call arguments in ExternalInterface before invoking registered callbacks

tschneidereit:
r? @yurydelendik

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2274)

### DIFF
--- a/src/flash/external/ExternalInterface.ts
+++ b/src/flash/external/ExternalInterface.ts
@@ -81,7 +81,19 @@ module Shumway.AVMX.AS.flash.external {
 
       this._addCallback(functionName, function (request: string, args: any[]) {
         var returnAsJS: Boolean = true;
-        if (args == null) {
+        if (args) {
+          var wrappedArgs = [];
+          for (var i = 0; i < args.length; i++) {
+            var arg = args[i];
+            // Objects have to be converted into proper AS objects in the current security domain.
+            if (typeof arg === 'object' && arg) {
+              wrappedArgs.push(self.sec.createObjectFromJS(arg, true));
+            } else {
+              wrappedArgs.push(arg);
+            }
+          }
+          args = wrappedArgs;
+        } else {
           var xml = this.convertToXML(request);
           var returnTypeAttr = xml.attribute('returntype');
           returnAsJS = returnTypeAttr && returnTypeAttr._value == 'javascript';


tschneidereit:
Review ping.

yurydelendik:
Looks good

---

Review status: 0 of 1 files reviewed, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/flash/external/ExternalInterface.ts, line 89 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2274#-JrT8agKNHqOAhyhN8CH)** ([raw file](https://github.com/mozilla/shumway/blob/68f34e03c96bf44340fa2f985ba85d4fb193731c/src/flash/external/ExternalInterface.ts#L89)):</sup>
(I don't think it would be a case here but) can you compare for not-null using `arg !== null` to avoid potential valueOf call.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2274)

yurydelendik:
Review status: all files reviewed, 1 unresolved discussion, all commit checks successful.
Reviewed files:
- src/flash/external/ExternalInterface.ts @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2274)

yurydelendik:
Review status: all files reviewed, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/flash/external/ExternalInterface.ts, line 84 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2274#-JrTAQMohDrU3CyotQwQ)** ([raw file](https://github.com/mozilla/shumway/blob/68f34e03c96bf44340fa2f985ba85d4fb193731c/src/flash/external/ExternalInterface.ts#L84)):</sup>
on second thought, also keep this as `(args != null)`

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2274)

tschneidereit:
Thanks for the review. I opted not to make the changes you requested, for the reasons given below.

---

Review status: all files reviewed, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/flash/external/ExternalInterface.ts, line 84 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2274#-JrTAQMohDrU3CyotQwQ)** ([raw file](https://github.com/mozilla/shumway/blob/68f34e03c96bf44340fa2f985ba85d4fb193731c/src/flash/external/ExternalInterface.ts#L84)):</sup>
Same as for the other check: a `valueOf` call can't happen here, making the check equivalent with what you propose for all intents and purposes.

---

<sup>**[src/flash/external/ExternalInterface.ts, line 89 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2274#-JrT8agKNHqOAhyhN8CH)** ([raw file](https://github.com/mozilla/shumway/blob/68f34e03c96bf44340fa2f985ba85d4fb193731c/src/flash/external/ExternalInterface.ts#L89)):</sup>
A `valueOf` can't happen here: objects are always truthy, without any further checks. (The only exception is `document.all`, but that doesn't apply here, obviously.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2274)

PR #2276, opened 2015-06-05
Support negative leading and letter-spacing in text fields.

tobytailor:
r? @tschneidereit

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2276)


tschneidereit:
Looks good, r=me with feedback addressed.

---

Review status: all files reviewed, 3 unresolved discussions, all commit checks successful.
Reviewed files:
- src/flash/text/TextField.ts @ r1
- src/flash/text/TextFormat.ts @ r1
- src/gfx/renderables/renderables.ts @ r1
- src/TextContent.ts @ r2
- test/swfs/acid/acid-text-4.fla @ r2
- test/swfs/acid/acid-text-4.swf @ r2
- test/test_manifest.json @ r2

---

<sup>**[src/flash/text/TextFormat.ts, line 34 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2276#-JrTKcTzJBJsIAqY0qtm)** ([raw file](https://github.com/mozilla/shumway/blob/5f0c24269c1f234ae88f161346a15177ababe9fb/src/flash/text/TextFormat.ts#L34)):</sup>
This and the other `_` additions are almost certainly invalid: the setters do coercions, which is why we're calling them. If the `TextFormat` ctor doesn't in fact coerce the values, then ideally we should test this. ( Actually, we should test it either way, but if you revert the change you're off the hook for it :) )

---

<sup>**[src/TextContent.ts, line 594 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2276#-JrTK6ddWo6MSnIug6FH)** ([raw file](https://github.com/mozilla/shumway/blob/5f0c24269c1f234ae88f161346a15177ababe9fb/src/TextContent.ts#L594)):</sup>
Modify the comment to include the new reason for skipping the steps.

---

<sup>**[src/TextContent.ts, line 633 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2276#-JrTJV-nZeBZZ8HZ53MB)** ([raw file](https://github.com/mozilla/shumway/blob/5f0c24269c1f234ae88f161346a15177ababe9fb/src/TextContent.ts#L633)):</sup>
Please add a comment explaining what this is doing. (It eliminates empty runs, I know. But it's not immediately self-explanatory [and I had to think for a moment to understand it], so should get a comment.)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2276)

PR #2279, opened 2015-06-09
Fixes WeakList to account for newly added keys

yurydelendik:
Fixes SWF 66 regressions (after 9343a3e): direct fix for WeakKey.forEach which does not account for newly inserted keys, hence does not run some scripts in the extension. 

Some scripts execute _advancableInstances.push during _advancableInstances.forEach operation. Question is: shall we allow such operation, i.e. mutate _advancableInstances during execution. If not, we shall probably remove forEach operation and replace it with getKeys, so behavior in the inspector and the extension would match.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2279)

### DIFF
--- a/src/base/utilities.ts
+++ b/src/base/utilities.ts
@@ -1512,10 +1512,12 @@ module Shumway {
 
   export class WeakList<T extends IReferenceCountable> {
     private _map: WeakMap<T, T>;
+    private _newAdditions: Array<Array<T>>;
     private _list: T [];
     constructor() {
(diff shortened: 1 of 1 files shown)


tobytailor:
Review status: :shipit: all files reviewed, all discussions resolved, all commit checks successful.
Reviewed files:
- src/base/utilities.ts @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2279)

PR #2283, opened 2015-06-11
Disables SSL server for reftest.

yurydelendik:
Trivial


tschneidereit:
Why is this change needed? If you skip reviewing, at *least* provide an explanation of why the PR exists.

yurydelendik:
it's conflicting with `grunt server` when both run at the same time

tschneidereit:
Makes sense.

PR #2288, opened 2015-06-12
Misc AVM1 fixes

yurydelendik:
Addresses top AVM1 errors from report:
- "Not Implemented AVM1MovieClip.attachAudio"
- "can't convert ("Cannot box: " + v) to primitive type"
- "x is undefined"
- "Date.UTC is not a function"

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2288)

### DIFF
--- a/src/avm1/lib/AVM1MovieClip.ts
+++ b/src/avm1/lib/AVM1MovieClip.ts
@@ -94,7 +94,14 @@ module Shumway.AVM1.Lib {
       return target != '/' ? prefix + target.replace(/\//g, '.') : prefix;
     }
 
-    public attachAudio(id) {
+    public attachAudio(id: any): void {
+      if (Shumway.isNullOrUndefined(id)) {
+        return; // ignoring all undefined objects, probably nothing to attach
+      }
+      if (id === false) {
+        return; // TODO stop playing all attached audio source (when implemented).
+      }
+      // TODO implement NetStream and Microphone objects to make this work.
       Debug.notImplemented('AVM1MovieClip.attachAudio');
     }
 
--- a/src/avm1/lib/AVM1MovieClip.ts
+++ b/src/avm1/lib/AVM1MovieClip.ts
@@ -149,21 +149,46 @@ module Shumway.AVM1.Lib {
       return as2mc;
     }
 
-    public beginFill(color, alpha) {
-      this.graphics.beginFill(color, alpha);
+    public beginFill(color: number, alpha: number = 100): void {
+      color = alToInt32(this.context, color);
+      alpha = alToNumber(this.context, alpha);
+      this.graphics.beginFill(color, alpha / 100.0);
     }
 
-    public beginBitmapFill(bmp, matrix, repeat, smoothing) {
-      if (!(bmp instanceof flash.display.BitmapData))
-      {
-        return;
+    public beginBitmapFill(bmp: AVM1BitmapData, matrix: AVM1Object = null,
+                           repeat: boolean = false, smoothing: boolean = false): void {
+      if (!alInstanceOf(this.context, bmp, this.context.globals.BitmapData)) {
+        return; // skipping operation if first parameter is not a BitmapData.
       }
+      var bmpNative = toAS3BitmapData(bmp);
+      var matrixNative = Shumway.isNullOrUndefined(matrix) ? null : toAS3Matrix(matrix);
+      repeat = alToBoolean(this.context, repeat);
+      smoothing = alToBoolean(this.context, smoothing);
 
-      this.graphics.beginBitmapFill(bmp, matrix, repeat, smoothing);
+      this.graphics.beginBitmapFill(bmpNative, matrixNative, repeat, smoothing);
     }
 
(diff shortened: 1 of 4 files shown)


tschneidereit:
Looks good. I'd like to see another version with my comments addressed, though, if only so I understand what's going on with the stuff I asked questions about.

---

Review status: all files reviewed, 5 unresolved discussions, all commit checks successful.
Reviewed files:
- src/avm1/lib/AVM1MovieClip.ts @ r1
- src/avm1/natives.ts @ r1
- src/shell/microTaskQueue.ts @ r2

---

<sup>**[src/avm1/lib/AVM1MovieClip.ts, line 420 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2288#-JrnMtVURVclcVMP36Td)** ([raw file](https://github.com/mozilla/shumway/blob/10feeaed25c26cb1736bf96b0c2fbf54b64a1cf2/src/avm1/lib/AVM1MovieClip.ts#L420)):</sup>
Does `isAVM1Instance` only exist on AVM1 MovieClip instances? If not, how is this check enough? Or does this method also work for Bitmaps? If so, the comment below should not only mention `movie`, but something like "AVM1 display object".

---

<sup>**[src/avm1/natives.ts, line 664 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2288#-JrnLc7fW8YIqSQ55rz3)** ([raw file](https://github.com/mozilla/shumway/blob/10feeaed25c26cb1736bf96b0c2fbf54b64a1cf2/src/avm1/natives.ts#L664)):</sup>
Please add a comment explaining what this function is for. Or maybe slightly more explanatory name would be enough. Perhaps `mapAVM1ListToJSArray`?

---

<sup>**[src/avm1/natives.ts, line 668 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2288#-JrnKfAZ0pAecFp4gMRc)** ([raw file](https://github.com/mozilla/shumway/blob/10feeaed25c26cb1736bf96b0c2fbf54b64a1cf2/src/avm1/natives.ts#L668)):</sup>
Nit: "array-like". Also, "This method is generic ..." sounds slightly better to me. Up to you, though.

---

<sup>**[src/avm1/natives.ts, line 669 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2288#-JrnLMT5RpFOJT-Npwua)** ([raw file](https://github.com/mozilla/shumway/blob/10feeaed25c26cb1736bf96b0c2fbf54b64a1cf2/src/avm1/natives.ts#L669)):</sup>
The first part of this seems unnecessary: `alIsArrayObject` subsumes it. (Also, preexisting, I'd much prefer `alIsArrayObject` to be `alIsArrayLike` instead. "ArrayObject" sounds like it's checking for an object of type `Array`.)

---

<sup>**[src/avm1/natives.ts, line 670 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2288#-JrnKwsZ-Ck1eALu6zWT)** ([raw file](https://github.com/mozilla/shumway/blob/10feeaed25c26cb1736bf96b0c2fbf54b64a1cf2/src/avm1/natives.ts#L670)):</sup>
Is this an error that can ever validly occur? If so, what does Flash do with it? I'm sure it doesn't throw this error. If not, this should assert instead of throwing a JS error. (If the error is caught by the interpreter, add a comment explaining that.)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2288)

yurydelendik:

---

<sup>**[src/avm1/natives.ts, line 670 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2288#-JrnKwsZ-Ck1eALu6zWT)** ([raw file](https://github.com/mozilla/shumway/blob/10feeaed25c26cb1736bf96b0c2fbf54b64a1cf2/src/avm1/natives.ts#L670)):</sup>
I did not address proper handling try/catch things in AVM1 yet. try/catch is not used a lot, so in most of the cases it will fail on any JS throw with diagnostic message in the interpreter for now. I would like to keep it this way (throwing JS errors), at least when enough tests for try/catch are passing. Which might also mean we have to re-test every AVM1 display object API catch with try/catch too, but all the above is out of scope of this PR.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2288)

yurydelendik:
> TEST-UNEXPECTED-FAIL | stas acid-morph-stas  | in firefox | trace of 1 != reference trace

Intermittent (fixing by #2290)

tschneidereit:
Thanks, looks great. r=me with that last tiny nit addressed.

---

Reviewed 3 of 3 files at r3.

---

<sup>**[src/avm1/natives.ts, line 666 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2288#-Jrs1zsxVfM_pgJSh9YT)** ([raw file](https://github.com/mozilla/shumway/blob/b266fa87f2f3bd11b67e67ef642f51511bcfbca8/src/avm1/natives.ts#L666)):</sup>
s/array-list/array-like/

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2288)

PR #2293, opened 2015-06-17
Add stub implementation of ExternalInterface to shell and inspector player

tschneidereit:


yurydelendik:
I don't see why this needed -- the SWF movie in most of the cases check for presence of ExternalInterface and properly handle the exception.

yurydelendik:
If it's for mocking the ExternalInterface, could you add it as pref'd off options?

tschneidereit:
> I don't see why this needed -- the SWF movie in most of the cases check for presence of ExternalInterface and properly handle the exception.

That's the thing: this is just not true, sadly. In many, many cases, SWFs just assume that ExternalInterface is available.

> If it's for mocking the ExternalInterface, could you add it as pref'd off options?

What's the advantage of having it pref'd off? When do you ever want the inspector or shell to die because of this instead of just logging a warning?

yurydelendik:
> When do you ever want the inspector or shell to die because of this instead of just logging a warning?

In default environment, ExternalInterface.available returns false and fails on methods calls. So FP will fail as well. Mocking the external interface, e.g. returning undefined on `document.location.href` call without raising exception might send logic to the different path or trigger unrelated failures for other properly designed SWFs.

tschneidereit:
> In default environment, ExternalInterface.available returns false and fails on methods calls. So FP will fail as well (you can notice that when debug build of FP is used). Mocking the external interface, e.g. returning undefined on document.location.href call without raising exception, might send logic to the different path or trigger unrelated failures for other properly designed SWFs.

That is a good point to some extent. However, we have 127 exceptions in the ads report that're caused by SWFs not checking for availability of ExternalInterface. I'd bet a serious sum on this causing more problems than the scenario you're worried about. Do you have any idea for how to get the best of both worlds? If not, I think we should make the behavior introduced in this PR the default, and I'll add an option for turning it off.

yurydelendik:
Looks good with option to turn it on added -- test runner might turn that option from command line.

---

Reviewed 7 of 7 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2293)

tobytailor:
@tschneidereit please rebase and land if we are still interesting in having this.

PR #2295, opened 2015-06-17
Catch errors in event handlers per target and properly continue execution. r=tobytailor

tschneidereit:
`EventDispatcher#dispatchEvent` aborts all event dispatching *for a specific target* upon unhandled exceptions in a handler. It logs a warning and then continues orderly execution. That includes dispatching the same event on other targets in the same, and in other phases. I.e., if a capturing phase handler throws, other capture phase handlers are run and the target and bubbling phases continue as expected. The included tracetest covers all these cases by throwing in all handlers.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2295)


tobytailor:
Reviewed 6 of 6 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2295)

PR #2297, opened 2015-06-18
Removes shumway.userInputSecurity option.

yurydelendik:
The nsIDOMWindowUtils.isHandlingUserInput shall be called in the same process where event was fired.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2297)

### DIFF
--- a/extension/firefox/chrome/ShumwayCom.jsm
+++ b/extension/firefox/chrome/ShumwayCom.jsm
@@ -175,6 +175,8 @@ var ShumwayCom = {
         // property and for usage as a sandbox for cloneInto operations.
         var gfxContent = gfxWindow.contentWindow.wrappedJSObject;
         ShumwayCom.createGfxAdapter(gfxContent, callbacks, hooks);
+
+        setupUserInput(gfxWindow.contentWindow, callbacks);
       },
 
       setupPlayerComBridge: function (playerWindow) {
@@ -194,10 +196,6 @@ var ShumwayCom = {
     // Exposing ShumwayCom object/adapter to the unprivileged content -- setting
     // up Xray wrappers.
     var wrapped = {
-      userInput: function () {
-        callbacks.sendMessage('userInput', null, true);
-      },
-
       setFullscreen: function (value) {
         value = !!value;
         callbacks.sendMessage('setFullscreen', value, false);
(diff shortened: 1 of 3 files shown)


tschneidereit:
Hmm, can't we do the entire user input detection in privileged code? It seems somewhat convoluted to let untrusted code listen to events, then tell chrome code that user input happened, which chrome code then has to verify. I suppose changing to that would be somewhat out of scope for this bug, but would be much cleaner.

Up to you whether you want to increase the scope or not. At least please file a follow-up bug to change to that, if there aren't any fundamental problems with it. In any case, r=me with nits addressed.

---

Reviewed 3 of 3 files at r1.

---

<sup>**[extension/firefox/chrome/content.js, line 49 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2297#-Js7lrvSDNVen8WsQ_Rn)** ([raw file](https://github.com/mozilla/shumway/blob/ef36c9876e9cf6bcd2c199a70d64fff470e5c1ec/extension/firefox/chrome/content.js#L49)):</sup>
I think we should memoize this. It's good not to get the interface until it's required, but if it is, we shouldn't pay the price for getting the interface every time.

---

<sup>**[extension/firefox/chrome/viewerWrapper.js, line 144 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2297#-Js7m7KaPZ0oBhD16sjc)** ([raw file](https://github.com/mozilla/shumway/blob/ef36c9876e9cf6bcd2c199a70d64fff470e5c1ec/extension/firefox/chrome/viewerWrapper.js#L144)):</sup>
Same note about memoizing the interface here.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2297)

yurydelendik:
Okay, removed events from gfx content page and moved them to the ShumwayCom. r=?

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[extension/firefox/chrome/ShumwayCom.jsm, line 443 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2297#-JsBHjE44QJ5e2VtHDf5)** ([raw file](https://github.com/mozilla/shumway/blob/9daad0d8605def734f099aee436231430b49d8dd/extension/firefox/chrome/ShumwayCom.jsm#L443)):</sup>
I explicitly provided 4th argument (I think it's false by default) -- just wanted annotate the decision to ignore the unsafe events.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2297)

tschneidereit:
Man, this is *so* much nicer, thank you! r=me.

---

Reviewed 5 of 5 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2297)

PR #2298, opened 2015-06-18
Bug 1176108: Use AMF encoding for local SharedObject data instead of JSON

tschneidereit:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2298)

### DIFF
--- a/src/avm2/natives/byteArray.ts
+++ b/src/avm2/natives/byteArray.ts
@@ -30,10 +30,15 @@ module Shumway.AVMX.AS {
   import assert = Shumway.Debug.assert;
 
   export module flash.net {
+    export enum AMFEncoding {
+      AMF0 = 0,
+      AMF3 = 3,
+      DEFAULT = 3
+    }
     export class ObjectEncoding extends ASObject {
-      public static AMF0 = 0;
-      public static AMF3 = 3;
-      public static DEFAULT = ObjectEncoding.AMF3;
+      public static AMF0 = AMFEncoding.AMF0;
+      public static AMF3 = AMFEncoding.AMF3;
+      public static DEFAULT = AMFEncoding.DEFAULT;
 
       static get dynamicPropertyWriter(): any /* flash.net.IDynamicPropertyWriter */ {
         notImplemented("public flash.net.ObjectEncoding::get dynamicPropertyWriter");
--- a/src/base/utilities.ts
+++ b/src/base/utilities.ts
@@ -794,11 +794,10 @@ module Shumway {
     }
 
     // https://gist.github.com/958841
-    export function base64ArrayBuffer(arrayBuffer: ArrayBuffer) {
+    export function base64EncodeBytes(bytes: Uint8Array) {
       var base64 = '';
       var encodings = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/';
 
-      var bytes = new Uint8Array(arrayBuffer);
       var byteLength = bytes.byteLength;
       var byteRemainder = byteLength % 3;
       var mainLength = byteLength - byteRemainder;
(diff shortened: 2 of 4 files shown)


tschneidereit:
Note that this doesn't contain a test because it's almost impossible to write an automated one. At least as long as we don't weakly hold the LSO data and introduce a `gc` function in the testing environment: as long as the original `SharedObject` instance is not GC'd, using `SharedObject.getLocal` with the same path always gives the same instance of the underlying data, so no matter how many times you call `SharedObject#flush`, you can't do before-after comparisons of the en-/decoding.

Code I used to test this manually:
```as3
import flash.net.SharedObject;
import flash.system.System;

var so:SharedObject = SharedObject.getLocal('dateTest');
var obj = {key: 'value', foo: true, bar: 2000, baz: 20.241, zorp: {}};
so.data.obj = obj;
so.flush();
so.close();
so = null;
System.gc();

so = SharedObject.getLocal('dateTest');
trace(so.data.obj === obj);
trace(JSON.stringify(so.data.obj) === JSON.stringify(obj));
```

yurydelendik:
Look good with comments addressed.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/base/utilities.ts, line 1178 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2298#-JsB6RtlSx7rpZnxQZ4u)** ([raw file](https://github.com/mozilla/shumway/blob/4122511a86978b99da343bb5003aea90990ec73c/src/base/utilities.ts#L1178)):</sup>
Assert for encoded.length be dividable by 4 above and use >> 2 instead here (or throw an exception)

---

<sup>**[src/flash/net/SharedObject.ts, line 64 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2298#-JsB5IcCxloVVO4SXnjM)** ([raw file](https://github.com/mozilla/shumway/blob/4122511a86978b99da343bb5003aea90990ec73c/src/flash/net/SharedObject.ts#L64)):</sup>
I wonder what logic that uses getDiskUsage does? If it does reset items when the result == 0, then this return value is somewhat incorrect.

---

<sup>**[src/flash/net/SharedObject.ts, line 176 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2298#-JsB5z1mY3gVRI_TFW_u)** ([raw file](https://github.com/mozilla/shumway/blob/4122511a86978b99da343bb5003aea90990ec73c/src/flash/net/SharedObject.ts#L176)):</sup>
We shall not add tests to the actual code. Move that to unit tests?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2298)

tschneidereit:
Looks like botio is doing some weird stuff here. Neither travis nor my local testing show the failing tracetest. Will investigate tomorrow.

---

Review status: 1 of 2 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---

<sup>**[src/flash/net/SharedObject.ts, line 64 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2298#-JsB5IcCxloVVO4SXnjM)** ([raw file](https://github.com/mozilla/shumway/blob/4122511a86978b99da343bb5003aea90990ec73c/src/flash/net/SharedObject.ts#L64)):</sup>
It's undocumented, so I'm not sure, but I *think* that it just returns the total amount of space LSOs for the given `url` take. It certainly doesn't modify storage, so I'm somewhat sure this is ok for now.

---

<sup>**[src/flash/net/SharedObject.ts, line 176 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2298#-JsB5z1mY3gVRI_TFW_u)** ([raw file](https://github.com/mozilla/shumway/blob/4122511a86978b99da343bb5003aea90990ec73c/src/flash/net/SharedObject.ts#L176)):</sup>
I'm not sure I agree. We have various places where we use somewhat complex asserts. Note that I didn't add this to make sure a scenario I thought of wouldn't regress. What I want to make sure is that we have a chance of getting to know about failures in scenarios I *didn't* think of. Which this assert gives us. (A chance, note, not a guarantee. We'd need release build reporting for that, which I don't think we should add.)

I'm going to leave this in, but if you feel strongly about it, let's discuss it.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2298)

PR #2299, opened 2015-06-19
Fixes various E4X-related issues that show up when running the Tamarin acceptance tests

tschneidereit:
And brings the number of uncaught VM-internal errors to below 10 and the uncaught exceptions to below 90. So that's pretty good.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2299)


tschneidereit:
I'll not ask for review this. The changes are not trivial by any means, but any reviewer would either have to treat them as trivial or spend a couple of days getting themselves up to speed on E4X and AVM2 internals. Both on the spec and implementation level. That won't happen and I'm *very* confident the changes are a strict improvement, so ...

PR #2301, opened 2015-06-20
Add tool for quickly sorting a list of SWF URLs into good/bad lists

tschneidereit:
Automatically stores state updates in localStorage, so use can be interrupted and continued seamlessly. Also, and importantly, doesn't allow network access or ExternalInterface. We don't support either in the shell, so for the purposes of verifying if something is a good ATS entry, this is useful. And safer than letting these insane SWFs go wild.

`<left arrow>` == good
`<right arrow>` == bad

Alerts when done and logs the list of good files. (Though they can also be found in the textarea with the label "Good").

@mbebenita, this should work for what we discussed. I don't have the list of known-ads, but @yurydelendik or @tobytailor should.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2301)

### DIFF
--- a/utils/swftinder/index.html
+++ b/utils/swftinder/index.html
@@ -1,0 +1,138 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>SWF Tinder</title>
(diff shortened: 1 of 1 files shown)


mbebenita:
Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2301)

PR #2312, opened 2015-07-10
Add Flash base param support to shuobject

Pneumaticat:
Shuobject before appeared to just be passing the document base as the flash base to the created iframe. Now, if set, it will pass the absolute URL version of params['base']; otherwise it'll just pass getDocumentBase() as it did before.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2312)


yurydelendik:
Looks good, thank you for the patch

PR #2314, opened 2015-07-14
Major SWF parser refactoring.

tobytailor:
This is a long overdue refactoring of the SWF parser. The code it now nicely formatted, fully typed and provides proper interfaces for all objects returned by the parser. Stream objects are now wrappers around a Uint8Array and a DataView object, with all custom reader methods implemented on the class' prototype instead of adding custom properties to a DataView instance. Furthermore we reduces memory usage of the parser by not storing unused fields and combining groups of boolean fields to single flag properties, accessible via provided enums/bitmasks.
 

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2314)


tschneidereit:
This looks pretty great! I added lots of comments, but most of them are quite minor.

It's very nice that this PR largely paves the way for the one major performance issue we still have in the parser: the two-phased parsing of shapes (and hence glyphs).

One change I'm not wild about is moving all tag handlers into the same package. I quite liked the `LowLevel` thing because it made clear which structures we intend to tell the non-parser world about and which not. (I do see that there's a fuzzy line between these categories, though, so I won't make an r+ dependent on reverting this.)

Is there any reason for not making the various interfaces defined in `SWFTags.ts` proper classes and `new`ing them? That make it easier to, in the future, splice in the classes we end up actually using in the player and thus skipping the whole conversion thing from one object to another.

---

Reviewed 21 of 21 files at r1.
Review status: all files reviewed at latest revision, 31 unresolved discussions, all commit checks successful.

---

<sup>**[src/base/SWFTags.ts, line 187 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO-7MmTx2UODTEOHJi)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L187)):</sup>
I'd much prefer `ColorTransform` for this. I know that the SWF spec uses`Cxform`, but that doesn't only look odd, it also requires far more contextual knowledge to grok.

---

<sup>**[src/base/SWFTags.ts, line 198 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO5rpyhtz9TUnGjZDO)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L198)):</sup>
I'd call this `SWFTag`.

---

<sup>**[src/base/SWFTags.ts, line 213 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO0GrgFxg--_F35dZR)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L213)):</sup>
Superfluous, exists on base class.

---

<sup>**[src/base/SWFTags.ts, line 270 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO6B0X1dZzHDXFSP36)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L270)):</sup>
`ClipEvents`?

---

<sup>**[src/base/SWFTags.ts, line 309 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO6hRs8VE62RWxfMXI)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L309)):</sup>
`preserveAlpha`

---

<sup>**[src/base/SWFTags.ts, line 341 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO7BUJImQXZNBrs71w)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L341)):</sup>
It's not terribly important (because SWFs usually don't contain a huge amount of buttons), but these four booleans could also be one `stateFlags` number.

---

<sup>**[src/base/SWFTags.ts, line 390 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOTdvED1aPxA5qFu4P)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L390)):</sup>
You could also change this to `export type Glyph = ShapeRecord[];` and eliminate the outer object (with some changes in other parsing code, of course). Which would be nice, because there are quite a few of them sometimes.

---

<sup>**[src/base/SWFTags.ts, line 474 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO9erLBVzMzM9GuKzC)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L474)):</sup>
Ideally, this'd also be an enum. Up to you, though.

---

<sup>**[src/base/SWFTags.ts, line 552 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuO8L1aJhdc_cjDLfX1)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L552)):</sup>
Can `isMorph` and `canHaveStrokes` be integrated into the `flags` field?

---

<sup>**[src/base/SWFTags.ts, line 609 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOA-GeCf1I9RVBjgv0)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L609)):</sup>
As long as this is an interface, it probably doesn't matter much, but the large amount of optional fields is a bit worrying here. Eventually we should clean that up, perhaps.

---

<sup>**[src/base/SWFTags.ts, line 610 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOANjhJPYwk1HgU4A9)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/base/SWFTags.ts#L610)):</sup>
Can this be an enum?

---

<sup>**[src/flash/display/MovieClip.ts, line 94 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOAsECwBhr9JPLz8nL)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/flash/display/MovieClip.ts#L94)):</sup>
It doesn't hurt, but you don't really need the boolean coercion here. The resulting code won't be any more efficient, or clear, I think.

---

<sup>**[src/flash/display/SimpleButton.ts, line 326 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOB57x7O8Leqf3qrs3)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/flash/display/SimpleButton.ts#L326)):</sup>
nit: missing whitespace after `:`. More importantly (and preexisting), is there any reason why we can't just define this default tag once and reuse the same instance? Something like `var defaultButtonStatePlacePbject`, defined above `ButtonSymbol`.

---

<sup>**[src/flash/text/TextField.ts, line 933 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOC5_5uU3xshSauDQH)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/flash/text/TextField.ts#L933)):</sup>
Nit: also wrap after `:`.

---

<sup>**[src/flash/text/TextField.ts, line 947 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOCElaP6-FaaJIvqnV)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/flash/text/TextField.ts#L947)):</sup>
We should eventually turn all of these into flags, too. Same for `TextField`. In another PR, though.
(Oh, and *here* I agree that the boolean coercion is required. Storing the masked number into a field is quite different from using it a few lines later.)

---

<sup>**[src/flash/text/TextField.ts, line 979 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOCo4cGHfqCR_iMqy-)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/flash/text/TextField.ts#L979)):</sup>
If you type the var here, you don't need to add the explicit cast a few lines below.

---

<sup>**[src/swf/parser/font.ts, line 106 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuODJTMKz9_8N89u9bJ)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/font.ts#L106)):</sup>
Ideally, we'd also just use the flags here, but I agree with limiting changes to the low-level parser for now.

---

<sup>**[src/swf/parser/image.ts, line 31 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuODlXRBu7Osjgeuefd)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/image.ts#L31)):</sup>
Wow.

---

<sup>**[src/swf/parser/module.ts, line 2 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOQ82AfBnPllSjE46v)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/module.ts#L2)):</sup>
2015

---

<sup>**[src/swf/parser/module.ts, line 69 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOQIyTHomVRvOfjlC3)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/module.ts#L69)):</sup>
Pretty sure these should be `255` or, even better (but no more correct) `0xff`. 
Or you could just turn them (and the offsets) into one `uint32`, of course ;)

---

<sup>**[src/swf/parser/module.ts, line 281 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuORv3dppqTOgQx8igq)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/module.ts#L281)):</sup>
This should always be assigned after `tag.depth`. That way, `tag.depth` ends up in the same slot for all instances, making the JIT's job easier.

---

<sup>**[src/swf/parser/module.ts, line 368 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuONRREfXGsA2JeJq3X)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/module.ts#L368)):</sup>
Can you replace these weird flag tests with simple enum checks?

---

<sup>**[src/swf/parser/module.ts, line 373 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuONbN-9RXLA5kfski5)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/module.ts#L373)):</sup>
As I said above, ideally the states would stay around as flags. I understand now that you want to contain the changes instead of extending them to the entire code base, so this should probably be a followup.

---

<sup>**[src/swf/parser/module.ts, line 415 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOSUZho5WRVG7UPiAI)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/module.ts#L415)):</sup>
Nit: type `tagEnd`

---

<sup>**[src/swf/parser/shape.ts, line 216 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOLP5QaA_lQFEOzKlY)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/parser/shape.ts#L216)):</sup>
Nit: no need to put braces around the `&` operations as long as you don't want to negate them.

---

<sup>**[src/swf/stream.ts, line 81 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOJ3xPPcoFl4ewRrEL)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/stream.ts#L81)):</sup>
I wonder what we should really be doing here? Throwing a string seems ... not right.

---

<sup>**[src/swf/stream.ts, line 103 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOM1s_PqDOIaeZiSfX)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/stream.ts#L103)):</sup>
This and perhaps `readSi8` would be excellent candidate for your closure compiler inlining annotation, I think.

---

<sup>**[src/swf/stream.ts, line 109 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOEmAECMI21GGrv8GE)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/stream.ts#L109)):</sup>
nit: missing `;`

---

<sup>**[src/swf/stream.ts, line 115 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOEnnmXEvc-FQ_MJgj)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/stream.ts#L115)):</sup>
nit: missing `;`

---

<sup>**[src/swf/stream.ts, line 127 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOE_Z4SA-6qD55823J)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/stream.ts#L127)):</sup>
nit: missing `;`

---

<sup>**[src/swf/SWFFile.ts, line 706 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2314#-JuOKTqNHTFfaDelfK-_)** ([raw file](https://github.com/mozilla/shumway/blob/442041765fb7050372dabbb0e7e26ad089aa9a4b/src/swf/SWFFile.ts#L706)):</sup>
There's really no reason to pass the tag in here instead of using it in the parser function itself.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2314)

PR #2320, opened 2015-07-29
Loosen allowDomain security check for cross-domain loading to only apply to SWFs

tschneidereit:
Currently we deny loading any resources from other domains if the loading SWF doesn't call `system.Security.allowDomain` with a string allowing the other domain. The idea is to avoid privilege escalations through the loaded SWF escaping from its sandbox and reaching into the parent SWF. This mitigation isn't required for images because those don't execute code anyway.

r? @yurydelendik

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2320)


tschneidereit:
Hum, this breaks the SWF on http://www.speakeasy.net/speedtest/ even more. I had hoped it'd work completely with this patch, as it fails due to the heuristic preventing image loading. However, it instead causes failures in loading the configuration XML file. Will investigate.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2320)

tschneidereit:
Ok, this now works as expected and makes the above-mentioned speedtest thing work. (It's dog-slow, though.)

yurydelendik:
We shall protect BitmapData.draw if remote origin image is painted (per http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/BitmapData.html#draw%28%29)

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2320)

tschneidereit:
Yes, definitely, but that's unrelated to this PR. I filed [bug 1188840](https://bugzil.la/1188840) for making sure that content can't access loaded images' data, and that's where we should do the `BitmapData#draw` work, too.

This PR is about loosening the protection of the *loading* SWF, not the *loaded* content, so `BitmapData#draw` doesn't enter the picture. Origin checks in `BitmapData#draw` aren't required right now because we only ever load data from other origins if a `crossdomain.xml` file allows it. We should change that, but again, that's [bug 1188840](https://bugzil.la/1188840). It took me an embarrassingly long time to figure out how very much unrelated the two are, at which point my patch got a lot smaller than it was before.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2320)

yurydelendik:
Looks good. Let's test

/botio test

---

Reviewed 5 of 5 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2320)

tschneidereit:
Thanks for the review!

---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2320)

PR #2332, opened 2015-08-11
Bug 1175070 - Fixes AVM1 ActionEquals2 behavior.

yurydelendik:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1175070

SWF spec defines ActionEquals2 behavior be the same as defined in ECMA-262 Section 11.9.3.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2332)


yurydelendik:
> Reference tests: FAILED
>    Trace log: http://areweflashyet.com:8081/aa3e54818ad5624/reftrace.log

Infra errors. See #2333 and [Bug 1193595](https://bugzilla.mozilla.org/show_bug.cgi?id=1193595)

r? @tschneidereit

tschneidereit:
I added quite a few comments below. In a way the most important change I'd request is to change the step comments. It's hard to compare the implementation with the spec if I have to translate the step names in my head all the time.

However, I also think that it'd make a lot of sense to at least do some things to speed this up. This isn't premature optimization because we know that this is going to be one of the hottest functions in the whole engine, sadly. (No, AS1/2 devs don't use strict equality all that often.)

(I also realized that we have a bug in `alToNumber`, because the empty string results in `0` in JS, not in `NaN`. Will send a PR to fix that.)

Cancelling review for now.

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

<sup>**[src/avm1/interpreter.ts, line 311 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWGSHUAHRIpbQIgEr5)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L311)):</sup>
Please add a link to the section of the spec.

---

<sup>**[src/avm1/interpreter.ts, line 312 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWJl00p9xUwBNXb9bC)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L312)):</sup>
Perhaps "Coerces x and y to the same type and returns true if they're equal, false otherwise."?

---

<sup>**[src/avm1/interpreter.ts, line 315 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWGXsxK4q1qnOcoCj5)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L315)):</sup>
By "steps 1-13" you mean step 1 with its sub-steps, right? If so, please change to "Step 1 can ...". As it is right now, the reader has to count out the sub-steps.
Also, please store the `typeof` results in vars at the beginning and use them everywhere below. The compiler *might* do it for you, but it'll at least take a bit, whereas it's straight-forward to do manually.

---

<sup>**[src/avm1/interpreter.ts, line 318 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWGrxEbVLmudjnU9Ml)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L318)):</sup>
Same here: please change these to "Steps 2,3", and also fix the numbering below.

---

<sup>**[src/avm1/interpreter.ts, line 322 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWHjXepGVCDA776RDL)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L322)):</sup>
Given how extremely hot this function is going to be, it might make sense to inline the relevant parts of `alToNumber` here and below. For this case, the inlining would be `return x === (y === '' ? (context.swfVersion < 5 ? 0 : NaN) : +y)`.

---

<sup>**[src/avm1/interpreter.ts, line 328 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWKa_30EkPcikwcovj)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L328)):</sup>
No need to use `alToNumber` here: it's 100% functionally identical to just using `+x`.
What's more, would `return +x === alToNumber(context, y)` not also be the same? It's be a huge load faster, that's for sure. Same in the next step.

---

<sup>**[test/swfs/avm1/operations/actionequals2.swf.trace, line 9 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWMVGn9Mwc-h6JISuo)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/test/swfs/avm1/operations/actionequals2.swf.trace#L9)):</sup>
Additional tests that'd make sense:
`false == {valueOf: function(){return false}}`
`true == {valueOf: function(){return true}}`
Also, similar things with `toString`.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2332)

yurydelendik:
Addressed almost of comments, except typeof optimization

---

Review status: 1 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[src/avm1/interpreter.ts, line 316 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWGXsxK4q1qnOcoCj5)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L316)):</sup>
No really, it's steps in the ecmascript 262 version 3 specification.


Per IRC, we agree to not cache `typeof x` just yet. However I can change that if needed.

---

<sup>**[src/avm1/interpreter.ts, line 326 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWHjXepGVCDA776RDL)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L326)):</sup>
actionequals2 is not present in swf < 5

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2332)

tschneidereit:
Very nice. r=me with the remaining feedback addressed. The `''` to `number` coercion being the most important. A test for that would be nice.

---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/avm1/interpreter.ts, line 311 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-Jwa6R_KlcdqnL9CMfNX)** ([raw file](https://github.com/mozilla/shumway/blob/c14e1fad427f002e1c81b3bd30c234ae0b520ac6/src/avm1/interpreter.ts#L311)):</sup>
Didn't see this last time around, but the header should contain the spec version: "EcmaScript 3, Section ..." (And no, the "Ecma" in "EcmaScript" isn't all upper-case, don't ask me.)

---

<sup>**[src/avm1/interpreter.ts, line 315 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWGXsxK4q1qnOcoCj5)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L315)):</sup>
Oh, ok. That makes me realize that the comment should contain the spec version.

---

<sup>**[src/avm1/interpreter.ts, line 322 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-JwWHjXepGVCDA776RDL)** ([raw file](https://github.com/mozilla/shumway/blob/90dfc339b095654beb9f58e57af196de3b5d3ba2/src/avm1/interpreter.ts#L322)):</sup>
That's excellent. It means that just using `+` is never correct, though: JS engines convert `''` to `0`, AVM1 to `NaN` :(

---

<sup>**[src/avm1/interpreter.ts, line 334 \[r2\]](https://reviewable.io:443/reviews/mozilla/shumway/2332#-Jwa7Jo_rRnz4tNQ_J4f)** ([raw file](https://github.com/mozilla/shumway/blob/c14e1fad427f002e1c81b3bd30c234ae0b520ac6/src/avm1/interpreter.ts#L334)):</sup>
Clever!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2332)

PR #2334, opened 2015-08-12
Bug 1148805 - Cache MovieClip child names in AVM1MovieClip for faster lookup

tschneidereit:
This patch reduces the time we spend under DisplayObject.performFrameNavigation during each turn of the event loop from ~1450ms to ~90ms on my machine, with the devtools profiler open. It's not the prettiest thing in the world, but also not so bad, and fairly small. See patch comment for details.

r? @yurydelendik.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2334)

### DIFF
--- a/src/avm1/lib/AVM1MovieClip.ts
+++ b/src/avm1/lib/AVM1MovieClip.ts
@@ -74,12 +74,20 @@ module Shumway.AVM1.Lib {
     }
 
     public initAVM1SymbolInstance(context: AVM1Context, as3Object: flash.display.MovieClip) {
+      this._childrenByName = Object.create(null);
       super.initAVM1SymbolInstance(context, as3Object);
       this._initEventsHandlers();
     }
 
     _lookupChildByName(name: string): AVM1Object {
-      name = alToString(this.context, name);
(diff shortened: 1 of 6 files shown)


yurydelendik:
Looks good.

---

Reviewed 6 of 6 files at r1.

---

<sup>**[src/avm1/lib/AVM1MovieClip.ts, line 211 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2334#-JwiUKFj78zbd385Ca4C)** ([raw file](https://github.com/mozilla/shumway/blob/8b13d88abc766dd84e5db8f025aa27dafbad9d99/src/avm1/lib/AVM1MovieClip.ts#L211)):</sup>
There is no comments in the introduced functions/code, but can we add comment why this variable was introduced, e.g. "This map helps to reduce the time we spend to resolve property in AVM1 (e.g. in event loop during DisplayObject.performFrameNavigation)"

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2334)

tschneidereit:
I added a comment explaining what the map is for and why it's useful to have.

PR #2337, opened 2015-08-12
Bug 1148805 - Do escaping of AVM1 property names before iteratively searching through scopes instead of doing it for each scope over and over again

tschneidereit:
While it fully works and gets some performance improvements, this is WIP-py in that it could be extended to other operations. f? @yurydelendik on the approach.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2337)


tschneidereit:
It might actually make sense to land this as-is: `_escapeProperty` completely vanishes from all profiles with it, whereas it takes close to 10% of execution time (excluding rendering) without it.

yurydelendik:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/avm1/interpreter.ts, line 885 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2337#-Jx5Wi7J_LeeLrMjVIpZ)** ([raw file](https://github.com/mozilla/shumway/blob/066bed51ad6710fbd1ff15925119cb6530b9cd66/src/avm1/interpreter.ts#L885)):</sup>
`variableName` can contain a path to the child, e.g.  '../a', './a/b:prop', 'a.b.__proto__' -- escaping must be happening to all its parts.

---

<sup>**[src/avm1/runtime.ts, line 598 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2337#-Jx5YuaLbGEbv41VK2A_)** ([raw file](https://github.com/mozilla/shumway/blob/066bed51ad6710fbd1ff15925119cb6530b9cd66/src/avm1/runtime.ts#L598)):</sup>
we don't need to `name.indexOf(ESCAPED_PROPERTY_PREFIX) !== 0` to avoid name collision when user will specify e.g. `__avm1__proto__`. (Which also mean we shall never call alToName twice on the same name -- did you have a case when we potentially do that?)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2337)

tschneidereit:
@yurydelendik, I removed the escaping and optimized a few other things. Makes the particle effects test case about 25% faster for me.

yurydelendik:
Review status: 0 of 14 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[src/avm1/interpreter.ts, line 954 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2337#-Jx5Wi7J_LeeLrMjVIpZ)** ([raw file](https://github.com/mozilla/shumway/blob/066bed51ad6710fbd1ff15925119cb6530b9cd66/src/avm1/interpreter.ts#L954)):</sup>
This comment was not addressed. We can remove it -- the avm1ResolveSimpleVariable logic will call alHasProperty which will normalize the name.

---

<sup>**[src/avm1/runtime.ts, line 66 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2337#-Jy9B927MKi7kmmNrEpA)** ([raw file](https://github.com/mozilla/shumway/blob/d7786c5e568c70a9e5128ac8467c037274584845/src/avm1/runtime.ts#L66)):</sup>
We can remove ESCAPED_PROPERTY_PREFIX

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2337)

tschneidereit:
@yurydelendik, I fixed the `SetVariable` issue we discussed yesterday in a follow-up patch in #2363.

PR #2339, opened 2015-08-12
Bug 1148805 - Only warn for valid but buggy AVM1 code if warning flag is set

tschneidereit:
The warnings can cause serious slowdowns, because they sometimes stringify complex objects, so they shouldn't happen in release builds.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2339)


yurydelendik:
looks good

/botio test

---

Reviewed 2 of 2 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2339)

PR #2340, opened 2015-08-14
Always allocate PreRenderState through pool, and ensure this won't cause further leaks

tschneidereit:
We had an ever so small leak that's fixed by this: http://i.imgur.com/nbBRzOT.png

(In case the image goes away: my dev browser instance used 53GB after letting a SWF of medium complexity run for a few hours.)

r? @yurydelendik

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2340)

### DIFF
--- a/src/gfx/nodes.ts
+++ b/src/gfx/nodes.ts
@@ -230,12 +230,14 @@ module Shumway.GFX {
 
   export class PreRenderState extends State {
     private static _dirtyStack: PreRenderState [] = [];
+    private static _doNotCallCtorDirectly = Object.create(null);
 
     matrix: Matrix = Matrix.createIdentity();
     depth: number = 0;
 
-    constructor() {
+    constructor(unlock: any) {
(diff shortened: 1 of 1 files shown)


yurydelendik:
Looks good.

Will it be better (from pattern view point) to always fail in the constructor and "allocate" using `var state = Object.create(PreRenderState); State.call(state);` ?

---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2340)

tschneidereit:
Thanks for the review.

I like the "ctor secret" pattern much better than manually creating the object, because it's very explicit both in the ctor itself and at all callsites. Thus, I'll leave this as is.

---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2340)

PR #2347, opened 2015-08-26
Clone buffers containing serialized GFX responses into player compartment before use

tschneidereit:
This fixes a truly bad performance fault in handling of `BitmapData` data.

r? @mbebenita

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2347)

### DIFF
--- a/src/base/dataBuffer.ts
+++ b/src/base/dataBuffer.ts
@@ -145,6 +145,22 @@ module Shumway.ArrayUtilities {
     }
 
     /**
+     * Clone the DataBuffer in a way that guarantees the underlying ArrayBuffer to be copied
+     * into an instance of the current global's ArrayBuffer.
+     *
+     * Important if the underlying buffer comes from a different global, in which case accessing
+     * its elements is excruiciatingly slow.
+     */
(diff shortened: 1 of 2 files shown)


mbebenita:
r+ with nits fixed.

---

Reviewed 2 of 2 files at r1.

---

<sup>**[src/base/dataBuffer.ts, line 155 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2347#-JxftRG9wo-mNBWTH6BG)** ([raw file](https://github.com/mozilla/shumway/blob/0beefe874506466ba567a9efd9b5d812c3ce75a8/src/base/dataBuffer.ts#L155)):</sup>
FromArrayBuffer doesn't clone all of the state in the current DataBuffer. Perhaps we should call this function something else. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2347)

tschneidereit:
> FromArrayBuffer doesn't clone all of the state in the current DataBuffer. Perhaps we should call this function something else.

I went the other way and just made it clone all of the state.

PR #2351, opened 2015-08-27
Report all "Not Implemented" messages when creating the SWF test run reports

tschneidereit:
rs=me

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2351)


tschneidereit:
We only have a few of those left, so seeing the long tail is useful.

PR #2355, opened 2015-08-28
Optimize color transforms and matrices for the common case.

mbebenita:
This reduces the size of updates by about 50% on some complicated animations. r? @yurydelendik

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2355)

### DIFF
--- a/src/base/dataBuffer.ts
+++ b/src/base/dataBuffer.ts
@@ -539,6 +539,24 @@ module Shumway.ArrayUtilities {
       }
     }
 
+    write2Floats(a: number, b: number): void {
+      var position = this._position;
+      this._ensureCapacity(position + 8);
+      this._requestViews(TypedArrayViewFlags.F32);
+      if (this._littleEndian === DataBuffer._nativeLittleEndian && (position & 0x3) === 0 && this._f32) {
+        this._f32[(position >> 2) + 0] = a;
+        this._f32[(position >> 2) + 1] = b;
+        position += 8;
+        this._position = position;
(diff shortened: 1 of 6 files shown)


tschneidereit:
Very nice! I added a few comments, but none are important, so I'll just merge this.

---

Reviewed 6 of 6 files at r1.

---

<sup>**[src/flash/geom/ColorTransform.ts, line 246 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2355#-JxttU9VPy9SP4dXdcdm)** ([raw file](https://github.com/mozilla/shumway/blob/3e3a2122cd9621e6b4e336a75e241f06f90697e0/src/flash/geom/ColorTransform.ts#L246)):</sup>
We should really just store - and transmit - color transforms as three uint32s. See [bug 1111836](https://bugzilla.mozilla.org/show_bug.cgi?id=1111836). That'd make this comparison so much faster, too.

---

<sup>**[src/player/remotingPlayer.ts, line 593 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2355#-JxtvBmhS_YxzweG0IT0)** ([raw file](https://github.com/mozilla/shumway/blob/3e3a2122cd9621e6b4e336a75e241f06f90697e0/src/player/remotingPlayer.ts#L593)):</sup>
These four lines could be moved below the `if/else` and removed from both blocks. Not important, though.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2355)

PR #2356, opened 2015-08-31
Adds AVM1 levels support (loadMovieNum).

yurydelendik:
Restores support of the levels and implements loadMovieNum and unloadMovieNum. Currently loading into _level0 is not supported yet -- requires refactoring of Loader to fully replace outermost SWF.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1035166 and partially https://bugzilla.mozilla.org/show_bug.cgi?id=1057555

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2356)


tschneidereit:
I think this isn't really the right approach. The `_levelN` paths are per AVM1-container. Multiple of those can exist in a single AVM2 instance, and all of them share the same `AVM1Movie` instance. So attaching them to the AVM2 `stage` isn't the right thing to do.

Consider an AVM2 SWF that creates a `Loader` and attaches it somewhere in its display list. In that loader, it loads an AVM1 SWF. That causes the loaded content to be wrapped inside an `AVM1Movie`. Inside that, the loaded content gets treated as `_level0` by the AVM1 instance. Now the AVM1 SWF loads another SWF into `_level1`. That other SWF must be in the same `AVM1Movie` container, to be shown at the same relative location in the AVM2 SWF. `AVM1Movie` is the container for a single AVM1 instance, which can contain multiple levels. [Bug 1035166](https://bugzilla.mozilla.org/show_bug.cgi?id=1035166) is largely about implementing this system.

Canceling review for now.

---

Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/avm1/lib/AVM1Globals.ts, line 402 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JyCV77_xQ7Reu5jQfaU)** ([raw file](https://github.com/mozilla/shumway/blob/39a980976792234b7994ce1a2e676c88988a4c1b/src/avm1/lib/AVM1Globals.ts#L402)):</sup>
I'm pretty sure levels can be multi-digit. I.e., you can load things into `_level11`. Just changing this to `substr(6)` should support that.

---

<sup>**[src/avm1/lib/AVM1Globals.ts, line 425 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JyCVa2oyxtpRj17uxLN)** ([raw file](https://github.com/mozilla/shumway/blob/39a980976792234b7994ce1a2e676c88988a4c1b/src/avm1/lib/AVM1Globals.ts#L425)):</sup>
Guard this with `release ||`.

---

<sup>**[src/avm1/lib/AVM1Globals.ts, line 621 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JyCW-eITwzwMg_mrwj-)** ([raw file](https://github.com/mozilla/shumway/blob/39a980976792234b7994ce1a2e676c88988a4c1b/src/avm1/lib/AVM1Globals.ts#L621)):</sup>
Guard this with `release ||`.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2356)

yurydelendik:
More refactoring:
- adds AVM1Movie root management methods and stores all roots at Sprite object (which is direct child of AVM1Movie)
- fixes incorrect logic around resolveLevelOrTarget (and removes AVM1Utils module)
- adds helper class for AVM1 movie clip loading.

/botio test

tschneidereit:
This has turned out very, very nice! It's much cleaner than before, in addition to being way more correct and all that. r=me with nits addressed.

---

Reviewed 5 of 13 files at r1, 1 of 10 files at r2, 13 of 14 files at r3.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

<sup>**[src/avm1/lib/AVM1MovieClip.ts, line 598 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JylcBFGNk6MALKz1ZCP)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/avm1/lib/AVM1MovieClip.ts#L598)):</sup>
Doesn't `setMask` with `null` remove the mask?

---

<sup>**[src/avm1/lib/AVM1MovieClipLoader.ts, line 100 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JyldgyZqhXUR80HEDzw)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/avm1/lib/AVM1MovieClipLoader.ts#L100)):</sup>
Can this be removed? If not, please add a `TODO` comment explaining why it's kept.

---

<sup>**[src/avm1/lib/AVM1Utils.ts, line 134 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JyleB6iFfNM81RQqMrg)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/avm1/lib/AVM1Utils.ts#L134)):</sup>
Please add a comment explaining what this does.

---

<sup>**[src/flash/display/AVM1Movie.ts, line 84 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JylecViwZDUOig_Iava)** ([raw file](https://github.com/mozilla/shumway/blob/65628cba42e1ef419f3a7cc801036c0f96e9c66e/src/flash/display/AVM1Movie.ts#L84)):</sup>
Why isn't this needed anymore? Please add a comment explaining why the hook isn't forwarded.

---

<sup>**[src/flash/display/AVM1Movie.ts, line 131 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-Jylelk4qs9IWVgaMEXC)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/flash/display/AVM1Movie.ts#L131)):</sup>
`_getRootForLevel`

---

<sup>**[src/flash/display/AVM1Movie.ts, line 132 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JylewlpYNghCq_DXOCH)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/flash/display/AVM1Movie.ts#L132)):</sup>
Why does `_level0` get special treatment here?

---

<sup>**[src/flash/display/AVM1Movie.ts, line 134 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-Jylf0a7uEIKWnp0rVuw)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/flash/display/AVM1Movie.ts#L134)):</sup>
When would this ever not be true? (It could be false if we supported removing `_level0` already, but right now we don't, so I don't see how. But even then ISTM that `getTimelineObjectAtDepth` should cover that.)

---

<sup>**[src/flash/display/AVM1Movie.ts, line 141 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JylavtGnbK3c8NAW1PC)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/flash/display/AVM1Movie.ts#L141)):</sup>
I'd much prefer for this to be called `_addLevel`, with the below `_removeLevel`. `setRoot` sounds like it sets the root of the receiver, and really, adding a level is exactly what this does.

---

<sup>**[src/flash/display/Loader.ts, line 754 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JylfmpTZKO3sdH3zI6e)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/flash/display/Loader.ts#L754)):</sup>
Ooh, annoying!

---

<sup>**[src/flash/display/Loader.ts, line 777 \[r3\]](https://reviewable.io:443/reviews/mozilla/shumway/2356#-JylfwbTULnlg6RjhAwl)** ([raw file](https://github.com/mozilla/shumway/blob/7ac9cb3830c7111706992e95499b4f0a667bc55c/src/flash/display/Loader.ts#L777)):</sup>
Why did you move setting `_level0` out of the ctor? Seems like we always want to immediately do that, so it'd be nice to keep it in there.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2356)

PR #2361, opened 2015-09-02
Update tsc to version 1.5.3

mbebenita:
This new version changes the semantics of enums. To get the old behavior, we need to use `const enums`. These however don't preserve names like enums used to, so we have to manually map enum values back to names. There is a TSC flag that does this, but I wanted to reduce the size of the generated code, so I only created maps where necessary. Most of the large maps are only used in once place and are gated with `release || ...` which will allows the closure compiler to eliminate them in release builds if they are not used.

The SWF module uses some enums as dictionaries, we should avoid this pattern since the enum values can't be constant folded. @tschneidereit 

r? @tschneidereit

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2361)


tschneidereit:
Nice. r=me with or without the one comment addressed. That can also be a follow-up.

---

Reviewed 51 of 51 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[src/avm2/abc/lazy.ts, line 58 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2361#-JyCO9hMfGNiFuAvFQhv)** ([raw file](https://github.com/mozilla/shumway/blob/dcfba0489c3f74945818ded39d51fbcba27c4cb5/src/avm2/abc/lazy.ts#L58)):</sup>
Would it make sense to guard the initialization by `release`, too? `var CONSTANTNames = release ? undefined : [..];`

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2361)

PR #2362, opened 2015-09-02
Cleanup

mbebenita:
Removes most of the unused code in utilities.ts, overall this is reduced from 138413 bytes to 85192 bytes. Functions that are not needed by the player are moved into gfx utilities.

Removes WebGL backend, not used and stale at this point.

Removes weird ABC file hash computation, not sure why we had that in there, especially since there is no code caching going on.

r? @tschneidereit

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2362)


tschneidereit:
Very nice! r=me

---

Reviewed 34 of 34 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2362)

PR #2367, opened 2015-09-02
Misc memory and timeline optimizations.

tobytailor:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2367)


src/flash/display/DisplayObject.ts
```diff
@@ -1852,12 +1852,19 @@ module Shumway.AVMX.AS.flash.display {
       release || assert(!symbol.dynamic);
       if (this._canHaveGraphics()) {
         release || assert(symbol instanceof flash.display.ShapeSymbol);
```
tschneidereit: While you're changing this, can you also swap the contents of the `assert` and the `if` for both this and the `TextSymbol` handling below? `instanceof` is cheaper than the `_canHaveGraphics` and `axIsType` checks.

tschneidereit:
Nice! r=me.

PR #2371, opened 2015-09-06
Fix E4X bugs around replacing and appending children, and normalization

tschneidereit:
Plus a sundry of other bugs I had to fix to avoid regressing tests that just so happened to pass because two bugs combined to make them work correctly, and I removed one of them.

Makes SWF Archive entry 1726 run successfully. There are pretty bad masking issues in the animation, but at least it fully runs through.

r=me because E4X.

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2371)


mbebenita:
> two bugs combined to make them work correctly, and I removed one of them.

I guess two wrongs make a right. :)

PR #2384, opened 2015-09-11
Adds the example of a simple swfm recorder.

yurydelendik:
[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2384)

### DIFF
--- a/examples/swfmplayer/recorder.html
+++ b/examples/swfmplayer/recorder.html
@@ -1,0 +1,94 @@
+<!DOCTYPE html>
+<!--
+Copyright 2015 Mozilla Foundation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


mbebenita:
Reviewed 3 of 3 files at r1.

---

<sup>**[web/iframe/viewerGfx.js, line 32 \[r1\]](https://reviewable.io:443/reviews/mozilla/shumway/2384#-Jz1nGTMj06m2yrWHZMA)** ([raw file](https://github.com/mozilla/shumway/blob/4581c2650117895698e704168d072def6479717e/web/iframe/viewerGfx.js#L32)):</sup>
We should have multiple recording limits, time, frames, and bytes I would think. But we should do that in a later PR.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2384)

mbebenita:

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/shumway/2384)

PR #2386, opened 2015-09-12
Remove dependency on tools in release bundles.

mbebenita:
This also changes swfm player to use closure compiled versions of the bundles.

While doing the refactoring, I noticed that the space twits example was stale and I removed it. Sorry for the sloppy PR, I could clean it up.

r? @yurydelendik

[Review on Reviewable](https://reviewable.io/reviews/mozilla/shumway/2386)


yurydelendik:
Thank you.

PR #2434, opened 2017-04-06
Fix windows build

ExE-Boss:
Fixes the `\r` symbol remaining at the end of `entry` in the file `utils/update-flash-refs.js:194`, which resulted in the build not working on Windows.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,3 +15,4 @@ Jeff Dyer
 Till Schneidereit <tschneidereit@mozilla.com>
 Jet Villegas <jet@mozilla.com>
 Claus Wahlers <cwahlers@mozilla.com>
+ExE Boss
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
That's not the only problem in windows build, see https://github.com/mozilla/shumway/pull/2442/files
The best way is just to specify list of files with strict LF

ExE-Boss:
That `\r` is put there by the Windows filesystem regardless of the value of `eol` in `.gitconfig`, so this patch is still also necessary (this only fixes the extension build).

ivanpopelyshev:
The idea is that you cant fix the code for `xargs` in the mozilla-build, so I have to add certain file in ".gitattributes" and prevent git-for-windows auto-CLRF mechanism.

There are many ways to fix that, I made a comment only to get your interest in the project again. Why did you try to build it?

ExE-Boss:
I tried to build it because I’m working on #2435.

Also, the purpose of this PR is just to make it possible for `grunt firefox` or any other `grunt` build command to complete successfully on Windows without needing to use WSL.

PR #2443, opened 2018-02-12
Fix PNG color type read

demurgos:
The color type is the byte at offset 9 in the IHDR PNG chunk
(offset 13 from the chunk size and 25 since the start of the file).
Shumway was incorrectly reading the byte 14 corresponding to the
compression method.

- See https://www.w3.org/TR/PNG/#11IHDR

### DIFF
--- a/src/swf/parser/image.ts
+++ b/src/swf/parser/image.ts
@@ -90,7 +90,7 @@ module Shumway.SWF.Parser {
     }
     image.width = readInt32(bytes, ihdrOffset + 4);
     image.height = readInt32(bytes, ihdrOffset + 8);
-    var type = bytes[ihdrOffset + 14];
+    var type = bytes[ihdrOffset + 13];
     image.hasAlpha = type === 4 || type === 6;
   }
 


ivanpopelyshev:
You can make that against https://github.com/ivanpopelyshev/shumway , it supports latest typescript version.

UPD. I've added this fix manually :)

demurgos:
Thanks for the reply. I noticed your project previously but it has a lot of changes and I haven't time to review it.

I am myself working on a player based off Shumway ([Open Flash](https://github.com/open-flash/open-flash)) but I have larger architectural changes (to support Rust/Wasm but also reuse only some parts for other tools). That's how I found the issue. Still, Shumway does not seem to use this info in the end because it resorts to the browser's parser and does not use the image size before the image is loaded.

ivanpopelyshev:
I have two branches there - one "master" thats just updated to latest typescript, and one WIP refactoring that i didnt finish yet. Its supposed to be new pixi plugin :)

